PR #5, opened 2018-06-19
first pass on the learn page

mansona:

PR #26, opened 2018-06-28
trying to fix width issue with learn page

mansona:
So I don't really understand what's going on with this but all I know is that it's not working correctly without changing how the padding works. @MelSumner let me know if you want me to detail more about this.

Before this change all of the content is super squashed into the middle of the screen:

screen shot 2018-06-28 at 10 24 11

After this change, the bounds of the "width" of the app are dealt with by the `.container` class: 

screen shot 2018-06-28 at 10 25 38

I wanted to use container just because I wanted to make sure that the navbar and the content were using the same measure to define their width.

### DIFF
--- a/app/styles/_learn.scss
+++ b/app/styles/_learn.scss
@@ -109,9 +109,12 @@
 //   padding: 0 0.5rem;
 // }
 
+main.container section {
+  padding: 1em 0;
+}
+
 //delete these after fixes are in ember-styleguide
 section {
-  padding: 1em 25em;
   h1, p {
     min-width: 100%;
   }
@@ -140,8 +143,7 @@ section {
     img {
       height: 50px;
       width: auto;
-    }  
+    }
   }
 
 }
-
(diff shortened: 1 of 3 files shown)


mansona:
This is actually a PR on https://github.com/ember-learn/ember-website/pull/5 so if you merge then it will be included in that PR 👍

app/styles/_learn.scss
```diff
@@ -109,9 +109,12 @@
 //   padding: 0 0.5rem;
 // }
 
+main.container section {
+  padding: 1em 0;
+}
+
 //delete these after fixes are in ember-styleguide
 section {
```
MelSumner: what happens when you set section to 100%?

app/templates/application.hbs
```diff
@@ -1,7 +1,9 @@
 {{#es-header}}
   {{es-navbar links=links}}
 {{/es-header}}
-<main>
-{{outlet}}
+
+<main class="container">
```
MelSumner: I'd like to try to avoid a class here if at all possible. but if we have to use one, how about something more specific, like .main-content or similar.

Also, can you add role="main" to this element.

mansona:
closing this in favour of #46

PR #79, opened 2018-10-23
Convert the Logos page https://emberjs.com/logos/

aklkv:
fixes: #16


app/styles/app.scss
```diff
@@ -474,3 +475,9 @@ body.tomster_faq #content {
 .cta-contribute {
   text-align: center;
 }
+
+// TODO: remove this one it's fixed in ember-styleguides
```
mansona: Can this be removed now with ember-styleguide v2.3.1 @aklkv ?

aklkv:
@mansona I have removed todo item but page still looks strange, my guess is that ember-styleguide have not release the new version yet

mansona:
@aklkv so it turns out it's a slightly different problem 😖 we're using a branch of the ember-styleguide because we want this site to use the new navbar. I have just rebased and pushed that branch but we should really think about getting the navbar fixed before we focus on anything else 🤔

PR #84, opened 2018-10-24
Team listing page

abhilashlr:


mansona:
I have a feeling that we might need to move over to the Ember data implementation for this because if you look at https://deploy-preview-84--ember-beta-website.netlify.com/team/ it won't load the data because it's looking for http://localhost:7784/data/team.json (which it really shouldn't)

We should not merge this until we solve the problem I think 🤔

abhilashlr:
I just noticed that this problem is persisting in learn page as well.

mansona:
This is blocked by #102

PR #86, opened 2018-10-26
upgrading ember-cli

aklkv:
This PR upgrades ember-cli to latest stable. One thing to note is that ember-tample-lint is omitted  on purpose. I will submit new PR with template-lint and fixes for it


.travis.yml
```diff
@@ -6,10 +6,6 @@ node_js:
 sudo: false
 dist: trusty
 
-branches:
```
mansona: can we add this back in? It will keep our Travis from slowing down too much 👍

package.json
```diff
@@ -3,40 +3,42 @@
   "version": "0.0.0",
   "private": true,
   "description": "Small description for ember-website goes here",
+  "repository": "",
```
mansona: can you add the repo please

.travis.yml
aklkv: sure

PR #102, opened 2018-10-29
Implement Ember Data for the data objects

mansona:
As part of this conversion, we noticed a few pages that rely on data files don't seem to render properly on the deployed Netlify application. This PR is an attempt to rectify these issues by using the data-access method that is known to work on the Ember Guides and other projects like this.

If this works in the Netlify preview app then I think we should consider using this method for all data access on the website 👍


app/models/showcase.js
```diff
@@ -0,0 +1,10 @@
+import DS from 'ember-data';
```
abhilashlr: Should we do `import attr from 'ember-data/attr'` instead?

locks: @abhilashlr we should follow the default recommendation.

mansona: Yea I think you're right @locks 👍 https://guides.emberjs.com/release/models/defining-models/#toc_defining-attributes

mansona:
So this PR is now ready for review. It fixes the issues we were having with the local/fastboot/prember environments being different and the production being broken 🎉 

There is one slight change/difference. The Showcase now is in a different order than it used to be. We can solve this in one of 3 ways: 

- add an "order" attribute and then sort by that order
- manually specify the ID and sort by that
- "hack" it and name the individual showcases `a-super-rentals`, `b-builds` and `c-api-docs` so that the *implied* order is correct

The first 2 options above require extra code to execute and the third one is a bit hacky but it should "just work". 

The alternative is that we don't currently care about the order and we can ship it as is: 

screen shot 2018-10-29 at 20 12 16

mansona:
This PR is now ready for review 👍 

The Learn page showcase projects are now in the correct order

PR #124, opened 2018-12-20
fixing images on ember-users page when running app locally

mansona:
There were some issues loading images on the ember-users page when running this app locally. I have a feeling we were working around a bug that was actually fixed upstream.

I want us to test the deploy-preview app to make sure this is working right before we merge 👍


mansona:
There seems to be an issue with the `{{asset-map}}` helper that is causing some inconsistencies in the Ember App, Fastboot and Prember environments 🤔 I am pretty sure it is related to https://github.com/RuslanZavacky/ember-cli-ifa/issues/32

The only way that I could think of solving this (apart from fixing this upstream) is to add a helper that always made sure that there was a leading `/` on image urls.

mansona:
@sivakumar-kailasam thanks for the suggestion 🎉 I hadn't even thought to do it that way until you mentioned it! 

I've done that now 👍

PR #125, opened 2018-12-20
Convert the team.yml into individual MD files and start using broccoli-static-site-json

mansona:
This PR is an extension of the work done in https://github.com/ember-learn/ember-website/pull/84 by @abhilashlr 

I had mentioned that #84 was "blocked by #102" but I wasn't very clear what I meant by that 🙈 I wanted to actually convert the PR to be using Ember-Data the same way that #102 was using it before merging the PR. This PR is that work 🎉 

I'm also streaming this right now (say hi 👋 ) so anyone interested in seeing the steps required can have a look at that video. I have also made sure that my Git history was close to the thought process and implementation process so you can follow along that way too 👍


mansona:
@sivakumar-kailasam thanks for pointing that out, I had intended to bring up that point in the PR but I just forgot 🙈 

The issue with sorting is that there is no real logic to the order in the previous implementation (i.e. the current website). If you can give me a rule I can sort the team members by that, but there is no rule currently and we would need to create one.

I had brought this up on my stream and some of the commenters mentioned that it might be good to sort alphabetically, but I opted not to do this in the PR because I didn't want to introduce it without discussion. Would you be happy with an alphabetical sort?

sivakumar-kailasam:
Afaict, it was based on the order members were added to each team on the main site. So in case we're moving away from that, might as well make the data be sorted alphabetically by name.

mansona:
@sivakumar-kailasam this has now been sorted correctly as we discussed in the Discord channel 👍 and is now ready for review and merge

Issue #119, opened 2018-12-04
Move deprecations from separate app into the main site


MelSumner:
## Goal 
Consolidate https://github.com/ember-learn/deprecation-app into this app. 

## To Do
- [ ] add the deprecations route as a child route to releases
- [ ] add ember-data, ember-cli and ember as child routes to deprecations
- [ ] each major version should have its own child route within the relevant project
- [ ] move over [deprecation content](https://github.com/ember-learn/deprecation-app/tree/master/content)

paxer:
Hi folks, I would like to help with this issue but I am relatively new to Ember community so pls expect many questions along the way :)

The first question - as I understood we need to move all content form this app https://github.com/ember-learn/deprecation-app to the new Ember site which is this repo (ember-website) right?

The first task is to add `deprecations` route as a child route to `releases` route as I understand? But when I check https://github.com/ember-learn/ember-website/blob/master/app/router.js - there is no  `releases` route there yet, should we create it as well or I am missing something?

paxer:
alright I think as with the old (current site) "Releases" is just a menu and should not be a route with URL, just wanted to make sure, but seems like this is what we want to leave in a new site as well

paxer:
A question regarding moving over deprecation content https://github.com/ember-learn/deprecation-app/tree/master/content

the content there is in Markdown format, should we move it as HTML or maybe leave it as Markdown  and use some Markdown to HTML component to render it on the page?

mansona:
Hey @paxer 👋 

Are you on the Ember Community Discord? can you ping me (@real_ate) either directly or in #dev-ember-learning to chat about this issue? 😄

paxer:
Hey @mansona 

I don't use Discord, not a problem to join though, however isn't it a good idea to keep all discussions about the issue in one place so anybody can review history in one place and not forced to jump between chat and GitHub?

mansona:
@paxer yes absolutely, we do tend to keep most of the conversation that is directly related to an issue or PR on GitHub because it's easier for new developers to discover 👍 In this particular case there is a bit of a more nuanced discussion that is going on and I thought you would be interested in having a bit more "real time" discussion 😄 

I'll get back to you with a response here in the next few days 👍

mansona:
Hey @paxer 👋 

[...]

PR #129, opened 2019-01-05
[WIP] Move deprecations from separate app into the main site 

paxer:
Fixes https://github.com/ember-learn/ember-website/issues/119


MelSumner:
Thank you for doing this. We have decided as team that we are only going to keep the deprecations app separate for _now_, because we need to work out some of the more challenging parts of making sure that URLs don't break in deprecation messages. Once we figure that out, we will come back to this! :+1:

Issue #115, opened 2018-11-25
Change the /tomster route to /mascots


MelSumner:
The /tomster/ route was already being redirected- the Ember version of the website should go ahead and use the /mascots/ route instead. 

To-Do:
- [ ] update the route name for /tomster/ to /mascots/
- [ ] update the subroutes
- [ ] update any tests
- [ ] ensure all tests pass and the project builds successfully

locks:
We need to redirect `/tomster/*`.

BradLeftley:
Hey I can pick this up 👍

locks:
@BradLeftley sounds good, thanks! Let us know here or on Discord if you have any problems.

MelSumner:
@BradLeftley is there anything I can help with to move this over the finish line?

locks:
Thanks!

PR #130, opened 2019-01-11
#115 Change tomster route to mascots

BradLeftley:


app/router.js
```diff
@@ -26,7 +26,8 @@ Router.map(function() {
   this.route('security');
   this.route('sponsors');
   this.route('team');
-  this.route('tomster', function() {
+  this.route('tomster', { path: '/*mascots' });
```
locks: This line should be removed as it is insufficient for redirecting, which will be taken care of by https://github.com/ember-learn/ember-website/pull/133.

PR #133, opened 2019-01-19
Port redirects from current website

locks:


.netlifyredirects
```diff
@@ -0,0 +1,6 @@
+/internal   /api
+/tomster    /mascots
+/tomster/*  /mascots/:splat
+/dasboard   /statusboard
+/guides     https://guides.emberjs.com/
+/guides/    https://guides.emberjs.com/
```
mansona: This needs to have "splatributes" added to the definition. You can see the documentation here: https://www.netlify.com/docs/redirects/#splats

.netlifyredirects
```diff
@@ -0,0 +1,6 @@
+/internal   /api
+/tomster    /mascots
+/tomster/*  /mascots/:splat
+/dasboard   /statusboard
+/guides     https://guides.emberjs.com/
```
mansona: you don't need both trailing and non trailing slashes in the definition: https://www.netlify.com/docs/redirects/#trailing-slash

locks: We do need splat and non-splat though, confirm?

.netlifyredirects
```diff
@@ -0,0 +1,4 @@
+/internal   /api
+/tomster/*  /mascots/:splat
+/dasboard   /statusboard
+/guides/*   https://guides.emberjs.com/:splat
```
mansona:
```suggestion
/guides/*   https://guides.emberjs.com/release/:splat
```

Looking at one of the examples of a "bad link" that this is supposed to fix https://emberjs.com/guides/models/ it seems like we need to include `release` in the redirect 🤔 

without this improvement https://deploy-preview-133--ember-beta-website.netlify.com/guides/models/ redirects to https://guides.emberjs.com/models/ which is a 404

PR #154, opened 2019-02-14
Statusboard

mansona:
This is porting the [statusboard app](https://github.com/ember-learn/statusboard) to be a route of the main website app 👍


MelSumner:
We need this to be in the same app as the main site. Trying to eliminate as many of the separate apps as possible.

Issue #151, opened 2019-02-11
Make sure the nav bar closes and we scroll to top after navigation


mansona:
This should utilise https://emberjs.com/api/ember/3.7/classes/RouterService/events/routeDidChange?anchor=routeDidChange to close the navbar and scroll to the top of the page

There is a chance that we might need to do this in Ember Styleguide (because of the navbar service) but it might be possible now in the website directly.

Let us know if you want to take this on or if you have any questions :+1:

patocallaghan:
@mansona I can take this one

mansona:
@patocallaghan nice one 🎉 you're a legend

mansona:
@patocallaghan now that https://github.com/ember-learn/ember-styleguide/pull/142 is merged you can update the dependency on styleguide directly since it's a git dependency right now https://github.com/ember-learn/ember-website/blob/master/package.json#L59

thanks again for looking at this 🎉

PR #155, opened 2019-02-15
Close navbar dropdown after router transition

patocallaghan:
Fixes https://github.com/ember-learn/ember-website/issues/151

After each router transition we use the `navbar` service to close dropdown menu. This bumps `ember-styleguide` to bring the `navbar#closePopupMenu` method introduced in https://github.com/ember-learn/ember-styleguide/pull/142

navbar


patocallaghan:
So it likes CI is failing potentially due to a caching issue. It looks like Travis is using an older version without the `closePopupMenu`. Any ideas @mansona?

mansona:
@patocallaghan I think it's just Yarn being difficult 🙈 it should be using the right one because of the yarn.lock file.

I'll take a look

patocallaghan:
Yeah, the commit SHA for ember-styleguide is correct for current master 🤷‍♂️

patocallaghan:
@mansona I've updated to use the commit SHA in the package.json. Will see if that works

patocallaghan:
Thanks @mansona, I think once we get a new release version of ember-styleguide we can go back to Yarn. The README will probably need updating too as it currently references `yarn`.

Issue #148, opened 2019-02-07
Implement the Payment Page for Tomster Comissions


mansona:
This issue is to re-implement https://emberjs.com/tomster/payment/ as `/mascots/payment`

Make sure you implement the "payment sent" page too

**WARNING:** this needs to be implemented alongside the core-team. You **must** reach out to the Ember Core Learning team before starting this implementation. We recommend that you ask about this issue in #dev-team-learning in the Ember Community Discord

patocallaghan:
@mansona I'd like to take this 🤚

PR #160, opened 2019-02-18
Add Mascot payment and sent pages

patocallaghan:
Closes https://github.com/ember-learn/ember-website/issues/148.

Similar to the the commission pages I've brought over the HTML as is, and not tried to optimise it or improve it in any way. I made some minor tweaks to the existing `form` styling to improve spacing but that's about it.

One thing I also noticed was the existing commission and payment have some inline JS using [tomster/wufoo.js](https://github.com/emberjs/website/blob/master/source/javascripts/app/tomster/wufoo.js). This has not currently been ported over. I'm not sure what the strategy is there?

## Payment page
**Current live (left) vs migrated page (right)**
image

## Payment sent page
payment-sent


mansona:
@patocallaghan thanks for this 🎉 I think we will **not** port over that javascript for now and we will run a test payment through this setup once it is merged just to make sure we don't need those JS bits and bobs 👍

PR #165, opened 2019-02-18
Port Home Page 

aklkv:


app/templates/application.hbs
```diff
@@ -4,6 +4,8 @@
   {{/es-navbar}}
 {{/es-header}}
 
+{{job-board-banner}}
```
mansona: This goes in the index and not the application 👍 it only needs to show up on the **home** page

Issue #146, opened 2019-02-07
Make sure there is a Netlify redirect from `/zoey` to `/mascots`


mansona:
This exists on the current ember website and needs to be replicated here :+1:

Netlify redirects are defined here: https://github.com/ember-learn/ember-website/blob/master/.netlifyredirects

Also make sure to remove the `/zoey` route from the Ember routes definition

patocallaghan:
I'll take this one ✋

Issue #145, opened 2019-02-07
Make sure there is a Netlify redirect from `/about` to the homepage


mansona:
This is something that seems to exist on the existing website and needs to be replicated here. 

Netlify redirects are defined here: https://github.com/ember-learn/ember-website/blob/master/.netlifyredirects

Also make sure to remove the `/about` route from the Ember routes definition

patocallaghan:
I may as well take this one too as it's pretty much the same as https://github.com/ember-learn/ember-website/issues/146

mansona:
yes indeed @patocallaghan 🎉 thanks a million! I'll mark it as in progress

PR #167, opened 2019-02-20
Redirect /about and /zoey pages

patocallaghan:
Fixes https://github.com/ember-learn/ember-website/issues/145
Fixes https://github.com/ember-learn/ember-website/issues/146

* Removes `/about` and `/zoey` router entries
* Removes `about` and `zoey` routes/tests/templates
* Removes their `prember` entries
* Adds redirect from `/about` to `/` (https://deploy-preview-167--ember-beta-website.netlify.com/about)
* Adds redirect from `/zoey` to `/mascots` (https://deploy-preview-167--ember-beta-website.netlify.com/zoey)

Regarding the `/about` redirect, while I agree that the live site is currently doing this behaviour, I actually think it's a bug. Should we not instead redirect to the first entry under the `About` dropdown, i.e. `/team`? I'd imagine that would map more to users' intent.

**GIF of redirects**
netlify-redirect


mansona:
so I just pushed a fix to this PR that is required because of the original intent of the `/about` redirect

The issue with redirecting to `/team` or something like that is that it almost certainly doesn't match the expected behaviour for the user. I think the `ember-users` route used to be under about i.e. `/about/ember-users` but I think that has been moved quite some time ago in the main website and the redirect was never setup https://emberjs.com/about/ember-users

Thre **real** fix will come from using an SEO tool and finding backlinks to the Ember website and making sure that each backlink actually redirects to the correct place. This was more of a "for now" catch-all implementation 👍

Issue #171, opened 2019-02-21
Add the community survey banner to the new website


mansona:
You may have noticed the community survey banner on the currently live website :tada: 

As we hope to launch the new website while the community survey is still active we need to re-implement the banner for the site.

You can take a look at this code: https://github.com/ember-learn/guidemaker-ember-template/blob/master/addon/templates/components/ember-survey.hbs (with the corresponding CSS and component file) to see how we do it in an Ember app :+1:

patocallaghan:
✋ I'll take this as a precursor to doing #173

mansona:
thanks @patocallaghan 🎉

Issue #173, opened 2019-02-21
Port the current year's community survey page


mansona:
This task is to port the https://emberjs.com/ember-community-survey-2019/ page from the previous website. There are some specific requirements for this implementation that make it slightly more complex than a normal static html port: 

### Todo
- [ ] port the static html for the page
- [ ] make sure that the banner introduced in #171 does **not** show on this page
- [ ] create a banner that has the countdown until the end of the survey as exists on the current live site

patocallaghan:
✋ I'll take this one as it's related to https://github.com/ember-learn/ember-website/pull/177

PR #177, opened 2019-02-22
Add Ember Community Survey banner to site

patocallaghan:
Closes https://github.com/ember-learn/ember-website/issues/171

Ports the Ember Community Survey Banner from [the guidemaker-ember-template repo](https://github.com/ember-learn/guidemaker-ember-template/blob/master/addon/templates/components/ember-survey.hbs)

image

### DIFF
--- a/app/templates/components/ember-survey.hbs
+++ b/app/templates/components/ember-survey.hbs
@@ -1,0 +1,7 @@
+{{#if config.link}}
+  <div class="survey-reminder-wrapper">
+    <div class="survey-reminder-container">
+      Take the <a href={{config.link}} target="_blank" rel="noopener">{{config.title}}</a>
+    </div>
+  </div>
+{{/if}}
--- a/app/templates/components/ember-survey.hbs
+++ b/app/templates/components/ember-survey.hbs
@@ -1,7 +1,7 @@
-{{#if config.link}}
+{{#if (and config.route (not-eq router.currentRouteName config.route))}}
   <div class="survey-reminder-wrapper">
     <div class="survey-reminder-container">
-      Take the <a href={{config.link}} target="_blank" rel="noopener">{{config.title}}</a>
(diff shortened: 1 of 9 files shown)


patocallaghan:
I guess it's no harm giving some forethought about how we'd implement the "Hide on the current active survey" functionality outlined in https://github.com/ember-learn/ember-website/issues/173.

We don't have to implement them here but these are just some thoughts I'm having.

* Keep the show/logic internal to the component
* Use the router service to detect current route, i.e. `currentRouteName`, and show/hide whether it passes some condition
* Add the current survey route name to the `config.survey` object and use that in the conditional to decide whether to show/hide the route, `config.currentSurveyRouteName` or something. At least this way when it's updated annually the route will be updated too and you don't need to change anything in the component

app/templates/components/ember-survey.hbs
```diff
@@ -0,0 +1,7 @@
+{{#if config.link}}
+  <div class="survey-reminder-wrapper">
+    <div class="survey-reminder-container">
+      Take the <a href={{config.link}} target="_blank" rel="noopener">{{config.title}}</a>
```
mansona: This can actually be a simple link-to now that it is linking inside the current application 🎉

patocallaghan: @mansona Any thoughts on shipping these changes as-is and I can then refactor to use an internal route/link-to when I implement #173? Doing that just means I don't need to create a route/empty template for this PR.

patocallaghan: @mansona I've updated the PR to:

* Add a `ember-community-survey-2019` route/template
* Updated the environment survey config to have a `route` property instead of `link`
* Refactored `ember-survey` to use `config.route` in `link-to` and not show the banner on the currently configured survey route
* Added some rendering tests for `ember-survey`'s show/hide logic

PR #221, opened 2019-03-11
adding query params for api redirects

mansona:
It was unclear by the documentation that this was actually necessary 🤔 https://www.netlify.com/docs/redirects/#query-params

I will test this in the preview app and self merge if it fixes the problem


mansona:
Self merge to fix the API link redirect 👍

PR #228, opened 2019-03-14
initial commit of editions and octane landing pages

### LINKED ISSUE #229: Add editions and octane pages
### The goal

Add an editions top level page, and editions/octane page.

MelSumner:
first commit to add editions and editions/octane pages to website. Resolves #229


app/templates/editions/octane.hbs
```diff
@@ -0,0 +1,32 @@
...
+  <li><a href="https://github.com/ember-learn/guides-source/issues/394">Write Guides</a></li>
+  <li><a href="https://github.com/ember-learn/guides-source/issues/588">Review & Edit Guides</a></li>
+</ul>
+
+<h2>Frequently Asked Questions</h2>
+<details>
+  <summary>What Happened to Module Unification?</summary>
+  Ember's conventions for project layout and file naming are central to the developer experience. It's crucial that we get both the technical and ergonomic details right. I wanted to provide an update about Module Unification and our plans for the file structure in Ember Octane.
+  <br/> In short, we do not plan to ship Module Unification in Octane. 
+  <br/>Read the full statement about Module Unification on our blog.
```
acorncom: Probably worth linking to the blog post here?

MelSumner: Can't put an interactive element inside of another interactive element. That's why I re-worded the prose here.

tomdale: If we can’t put links inside a `details` element, we should probably use a non-interactive element instead. The inability to use hyperlinks is a big cost and I don’t think hiding the content by default is necessary here. People may want to skim the content without having to click a bunch of disclosure triangles.

MelSumner: Sounds good. Will change. 👍

josemarluedke: The part "I wanted to provide" reads a little weird. I would change that to "We wanted to provide". Because this is not a blog post, there is not a reference to an author in the page.

jenweber:
@MelSumner can we go ahead and ship it? Thanks for your work!

jenweber:
Once this is merged, please merge https://github.com/ember-learn/guides-source/pull/641 since it updates the URL used in the Octane Guides banner.

jenweber:
Merged! Thanks Mel for your work and others for the reviews.

Issue #233, opened 2019-03-16
Add `ember-a11y-testing` to this repository


jenweber:
It's one of Ember's goals to make sure that our sites meet accessibility standards. We have some work to do to make it as easy as we can to catch mistakes.

For this PR:
- install [ember-a11y-testing](https://github.com/ember-a11y/ember-a11y-testing)
- Write an acceptance test that visits each url in the website router, and calls the `a11yAudit` function (see the addon README for details)
- On failures, log a warning (instead of a hard failure) so that people can run this test suite and use it in their subsequent small PRs to fix the problems, rather than one "big bang" PR

You can learn more about ember-a11y and accessibility in general in [this Ember.js blog article](https://blog.emberjs.com/2018/06/17/ember-accessibility-and-a11y-tools.html)

icole:
I will try to take a look at this. Discussed this with @jenweber in person at the contributors workshop to get started

MelSumner:
Thank you for doing this work @icole !

PR #240, opened 2019-03-18
Add ember-a11y-testing with ability to toggle a11yAudit acceptance tests

icole:
This adds support for ember-a11y-testing with acceptance test coverage of all of the routes that was requested in #233. 

By default each acceptance test per route will not run to avoid having failing tests in CI. Once a particular route has been addressed and all a11yAudit checks pass then adding `a11y: true` to the pages hash will run it as a normal test that expects a passing audit

I had tried to get this to work using QUnit `todo` as the default state, but unfortunately it appears testem does not support `todo` https://github.com/testem/testem/pull/1101


tests/acceptance/visual-routes-test.js
```diff
@@ -0,0 +1,60 @@
...
+      await prev;
+
+      await visit(config.route);
+
+      await percySnapshot(config.title);
+    }, Promise.resolve());
+  });
+
+  pages.forEach(function(page) {
+    (page.a11y_audit ? test : todo)(`testing a11y audit for ${page.route}`, async function (assert) {
```
dbollinger: 👍

tests/acceptance/visual-routes-test.js
```diff
@@ -0,0 +1,62 @@
...
+import a11yAudit from 'ember-a11y-testing/test-support/audit';
+
+module('Acceptance | visual regression', function (hooks) {
+  setupApplicationTest(hooks);
+
+  // This is used to map the current **new** routes to what they used to be in the
+  // old ruby app. This is to allow us to compare what the new app looks like
+  // vs the old app.
+  const pages = [
+    { title: '/builds/', route: '/releases' },
```
jessica-jordan: Would it make sense to add a code comment here explaining the `a11y` switch? E.g.
```suggestion
    { title: '/builds/', route: '/releases'/*, a11y: true */ },  /* set the a11y property of a page object to true to run the audit for this particular route */
```

icole: I like this suggestion to add clarity. thanks!

icole:
Thanks @jessica-jordan! Applied your suggestion and rebased from upstream. Let me know if I can do anything else.

PR #252, opened 2019-03-21
fixing the svg fingerprinting

mansona:
We had an Ember User's image not rendering because they submitted their image as a SVG, this enables the use of SVGs without breaking the links 🎉


mansona:
This PR actually already fixed https://github.com/ember-learn/ember-website/pull/278 and we should not be overriding extensions directly, we should extend the predefined extensions 👍

PR #257, opened 2019-03-21
Update the MU section on the status board.

mixonic:
Update the MU statusboard entry


mansona:
@mixonic I think this is worth merging relatively soon 👍 if we need to change it later we can do that 😄

PR #277, opened 2019-03-22
Address didTransition deprecation

locks:
Things I have tried:

- `this.on('routeDidChange'` in EmberRouter's `init` - handler not called
- Inject the router service in EmberRouter and add the handler on init - max call stack reached
- Inject the router service in application route and add the handler on init - handler is not called

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -11,11 +11,12 @@ const Router = EmberRouter.extend({
   navbar: service(),
   metrics: service(),
   fastboot: service(),
+  router: service(),
 
   init() {
     this._super(...arguments);
 
-    this.on('routeDidChange', () => {
+    this.router.on('routeDidChange', () => {
       this.navbar.closePopupMenu();
       this._trackPage();
     })
--- a/app/router.js
+++ b/app/router.js
@@ -12,10 +12,13 @@ const Router = EmberRouter.extend({
   metrics: service(),
   fastboot: service(),
 
-  didTransition() {
+  init() {
     this._super(...arguments);
-    this.navbar.closePopupMenu();
-    this._trackPage();
+
+    this.on('routeDidChange', () => {
+      this.navbar.closePopupMenu();
+      this._trackPage();
+    })
   },
 
   _trackPage() {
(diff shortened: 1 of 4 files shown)


acorncom:
@locks looks like you have failing tests

app/router.js
```diff
@@ -6,16 +6,19 @@ import { get } from '@ember/object';
...
   metrics: service(),
   fastboot: service(),
 
-  didTransition() {
+  init() {
     this._super(...arguments);
-    this.navbar.closePopupMenu();
-    this._trackPage();
+
+    this.on('routeDidChange', () => {
```
rwjblue:
```suggestion
    this.router.on('routeDidChange', () => {
```

app/router.js
```diff
@@ -6,16 +6,19 @@ import { get } from '@ember/object';
 import { scheduleOnce } from '@ember/runloop';
 
 const Router = EmberRouter.extend({
-  navbar: service(),
   location: config.locationType,
   rootURL: config.rootURL,
+  navbar: service(),
   metrics: service(),
   fastboot: service(),
```
rwjblue:
```suggestion
  fastboot: service(),
  router: service(),
```

locks: I've tried this and I get an error:

```
Uncaught RangeError: Maximum call stack size exceeded at http://localhost:7357/assets/vendor.js, line 23173
```

rwjblue: Tweaked the suggestion...

locks: I can't the invocation right, and I don't know what I'm doing wrong 🤔

acorncom: @locks would upgrading beyond the Ember 3.5.x that we’re on here help?

mansona:
@locks @rwjblue is this way of doing router `routeDidChange` stuff in the **application** route recommended? I want to make sure that we're doing the most recommended thing here and make sure that we bring this documentation upstream to the right places (deprecations documentation or ember-metrics)

locks:
@mansona yes, and I've opened an issue for updating the deprecation guide. There is a dependency loop between `EmberRouter` and `RouterService` that prevents us from doing it in `EmberRouter`.

mansona:
@locks thanks a million for following up on this 🎉

PR #283, opened 2019-03-25
Remove out of date EmberConf 2019 callout

amyrlam:
WAS: EmberConf 2019 callout
IS: No EmberConf 2019 callout

See: `netlify/ember-website/deploy-preview` link below of homepage emberjs.com or Percy!

Note: We need to fix the other Percy's besides the homepage separately, was discussing with @locks on Discord, need to further sync with him.

### DIFF
--- a/app/styles/app.scss
+++ b/app/styles/app.scss
@@ -26,6 +26,18 @@
 
 @import 'layout';
 
+// any dynamic sections of the site need to be hidden in percy to prevent false
+// positives in the diffs
+@media only percy {
+  .releases-canary code {
+    visibility: hidden;
+  }
+
+  .highcharts-wrapper {
+    visibility: hidden;
+  }
+}
+
 /* Deprecating these colors/variables */
 $blue-color: #3c64b6;
 $black-color: #252525;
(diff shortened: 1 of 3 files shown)


app/styles/app.scss
```diff
@@ -26,6 +26,18 @@
 
 @import 'layout';
 
+// any dynamic sections of the site need to be hidden in percy to prevent false
+// positives in the diffs
+@media only percy {
```
amyrlam: @mansona thanks for doing this, I wasn't sure if team / @locks were cool with them being hidden in Percy or wanted the tests to not be dynamic instead. lgtm! 👍

locks: I'm down with whatever yinz decide 💪

mansona: @amyrlam @locks we are not able to test them in Percy because we cannot (currently) wait for HighCharts to be finished animating before taking the Percy snapshot.

All dynamic things should be hidden by default and if we want to add back in Percy checks later once we're able to do the work that's great 👍 we should not have long-running papercuts that make us want to turn off CI tools 😞

amyrlam: sounds good to me!

PR #311, opened 2019-04-13
Releases page: Fix release version to 3.9, add LTS dates for 3.8

lupestro:
I removed the 2.16 LTS from the bottom of the list, as I added the 3.8 LTS to the top of the list. Since 2.18 (now the bottom of the list) expired its final support date back in February, it is reasonable to infer that all prior LTS versions are no longer supported. I also bumped the data so that 3.9 is shown as the current latest release, rather than 3.8.

### DIFF
--- a/app/templates/releases/index.hbs
+++ b/app/templates/releases/index.hbs
@@ -46,6 +46,20 @@
   <tbody>
     <tr>
       <td>
+        3.8
+      </td>
+      <td>
+        Feb 2019
+      </td>
+      <td>
+        Sep 2019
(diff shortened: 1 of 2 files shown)


kategengler:
I would have expected the release data to come from here: https://github.com/ember-learn/ember-website/blob/master/data/project/ember/release.md @mansona?

mansona:
@kategengler no this is right 👍 I don't know **why** the landing page has opted to display the "path to 3.10" but this is the correct change to fix the issue

PR #320, opened 2019-05-15
Adds Ember Engines on ember ecosystem

villander:
Recently we added Ember Engines on Ember Guides - https://guides.emberjs.com/release/applications/ember-engines/

I think that very important put on ember website the reference to Ember Engines, because he is maintained by ember core team as well 

@jenweber @sivakumar-kailasam

### DIFF
--- a/app/templates/learn/index.hbs
+++ b/app/templates/learn/index.hbs
@@ -82,6 +82,10 @@
   <strong>FastBoot</strong>: Server-side rendering library for Ember applications. Check out the <a href="http://www.ember-fastboot.com/quickstart">Quickstart</a> to get up and running, and the <a href="http://www.ember-fastboot.com/docs/user-guide">User Guide</a> for more detailed information.
 </p>
 
+<p>
+  <strong>Ember Engines</strong>: Ember addon that allow multiple logical applications to be composed together into a single application from the user's perspective. Check out their <a href="http://ember-engines.com/guide">interactive documentation</a>.
+</p>
+
 <h2 id="showcase" class="anchorable-toc">Showcase</h2>
 <p>In this section you will find applications that are maintained by the Ember.js teams with the help of contributors. While software is always a work in progress, the goal is to showcase patterns and solutions applied in real-world applications.</p>
 <p>Whether you're simply interested in checking out how some feature is implemented, or you're looking to contribute, one of these projects might pique your interest!</p>
--- a/app/templates/learn/index.hbs
+++ b/app/templates/learn/index.hbs
@@ -83,7 +83,7 @@
 </p>
 
 <p>
-  <strong>Ember Engines</strong>: Ember addon that allow multiple logical applications to be composed together into a single application from the user's perspective. Check out their <a href="http://ember-engines.com/guide">interactive documentation</a>.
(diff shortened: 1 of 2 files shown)


app/templates/learn/index.hbs
```diff
@@ -82,6 +82,10 @@
   <strong>FastBoot</strong>: Server-side rendering library for Ember applications. Check out the <a href="http://www.ember-fastboot.com/quickstart">Quickstart</a> to get up and running, and the <a href="http://www.ember-fastboot.com/docs/user-guide">User Guide</a> for more detailed information.
 </p>
 
+<p>
+  <strong>Ember Engines</strong>: Ember addon that allow multiple logical applications to be composed together into a single application from the user's perspective. Check out the <a href="http://ember-engines.com/guide">Ember Engines Guides</a>.
```
jenweber:
```suggestion
  <strong>Ember Engines</strong>: Ember addon that allows multiple logical applications to be composed together into a single application from the user's perspective. Check out the <a href="http://ember-engines.com/guide">Ember Engines Guides</a>.
```

villander: Fixed! @jenweber

villander:
fixed @jenweber @locks

jenweber:
To change the description of the CLI team, we need their approval, which we are waiting for.

app/templates/team.hbs
```diff
@@ -51,7 +51,7 @@
   <p class="text-center">
     The Ember CLI core team is responsible for maintaining ember-cli, the command line interface for managing and packaging
     Ember.js applications and addons. The team also maintains many of the addons in the default blueprint as well as
-    fastboot.
+    Fastboot and Ember Engines.
```
Turbo87: while it might be true that some people on the CLI team maintain those addons together with other people, I'm not sure it's correct to say that "the CLI team" maintains them. I personally am not maintaining them or even using them, and don't want the additional pressure of having to maintain those addons too. Since they are not directly related to Ember CLI I'm not sure that the CLI team should even be responsible for them. If they are "official" addons then maybe we should have dedicated teams for them 🤔

dgeb: @Turbo87 I agree this is a misrepresentation where engines are concerned, and perhaps fastboot as well. Personally, I'm in the awkward position of helping to maintain engines but not currently using them, so I can relate to your perspective.

It's worth discussing whether these projects should have their own "official" teams and representation on this page. Certainly the CLI team should not be assigned more responsibility than they want.

villander: After this, I super agree with you guys! @dgeb @Turbo87

Fastboot & Engines are official projects of Ember core team, but the responsibility is not clear of wich team members are dedicated on this. Can be very cool we having  a "Ember Engines Core Team" & "Fastboot Core Team"

rwjblue: Ya, I don't think this particular edit is needed. Can we drop the changes to this file and land  the other change?

villander: yup @rwjblue, I'll do it

villander: done @rwjblue @jenweber

PR #345, opened 2019-06-24
Revert "Bump ember-styleguide to get new footer link"

sivakumar-kailasam:
Reverts ember-learn/ember-website#342

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3742,9 +3742,9 @@
       },
       "dependencies": {
         "rsvp": {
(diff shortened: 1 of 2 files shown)


MelSumner:
I don't think we need to revert it, since it's not what was causing the issue. Can you confirm @sivakumar-kailasam ? I'll close the PR if so.

sivakumar-kailasam:
Oh we don't need this anymore. I opened this to test the changes on netlify

PR #352, opened 2019-06-27
Alumnizing Stef from Framework Team

wifelette:
...but not actually adding him to archive since he's still on CLI as far as I know and I don't _think_ we want folks in multiple places? Maybe if in the future the alums section also listed what teams people were from :)


stefanpenner:
:shipit:

PR #359, opened 2019-07-07
Add Stale app to repo

amyrlam:
## What it does
See: https://probot.github.io/apps/stale/. I added the Stale app to ember-blog and ember-website (these are the two repos that I have the most context on). I used the defaults from the Stale website. 

Note: Stale bumps issues and pull requests.

Tagging @jenweber cause we talked about it at the meeting, and @hakilebara based on his interests re: Ember-Learn Open Pull Requests Statistics. I think that this bot could really help us with keeping the "backlog" up to date, even if it's a #wontfix. I will also share in the Learning Channel.

Also added to ember-blog: https://github.com/ember-learn/ember-blog/pull/179

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,0 +1,22 @@
+# Number of days of inactivity before an issue becomes stale
+daysUntilStale: 60
+
+# Number of days of inactivity before a stale issue is closed
+daysUntilClose: 7
+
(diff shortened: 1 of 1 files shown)


amyrlam:
Thanks @rwwagner90, we can see how this goes and move to more repos besides website and blog if good!

PR #360, opened 2019-07-09
Update stale settings

amyrlam:
- Make settings 2x longer for issues. If it's been open for 4mo, seems safe to push to project management tool e.g. GitHub Projects / Notion / other we are using? 
- Change label to `stale` vs. `wontfix` (I can do this manually as well, seems more accurate.)

If we are cool with this, I can make the same changes on ember-blog.

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -2,17 +2,13 @@ pulls:
   daysUntilStale: 60
   daysUntilClose: 7
   markComment: >
-    This pull request has been automatically marked as stale because it has not had
-    recent activity. It will be closed if no further activity occurs. Thank you
-    for your contributions.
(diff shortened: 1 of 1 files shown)


amyrlam:
For now, I removed `wontfix` labeling from PR's and issues, since is not accurate.

amyrlam:
Sounds good... could reword issues and PR text? 

Can contributors also reopen a stale issue or PR? Or is that not possible, so a new issue is preferred?

amyrlam:
Discussed in meeting that contributors cannot reopen issues / PRs.

Updated the copy for issues and PR's. Looking for a final review! Then can make the change on `ember-blog`.

amyrlam:
Doing same on ember-blog https://github.com/ember-learn/ember-blog/pull/183, seems like can see how it goes on these two for now!

PR #375, opened 2019-08-07
Fix mascots

mansona:
This was an attempt to fix the mascots page using just the css helpers available in the styleguide. As you can see from the inclusion of some CSS it didn't quite hit that goal 😞 

We should discuss ways to make this less necessary or if we will, in fact, need some custom CSS for this page

https://deploy-preview-375--ember-website.netlify.com/mascots/

mascots

### DIFF
--- a/app/components/tomster-list.js
+++ b/app/components/tomster-list.js
@@ -1,0 +1,18 @@
+import Component from '@ember/component';
+import { computed } from '@ember/object';
+import { sort } from '@ember/object/computed';
+
+export default Component.extend({
+  tagName: '',
+
+  sorting: Object.freeze(['date:desc']),
+  sortedModel: sort('filteredModel', 'sorting'),
+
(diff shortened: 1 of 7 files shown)


mansona:
I have used the latest style guide and we are now able to use significantly less CSS in this PR but we still have a little way to go 👍

MelSumner:
Let's talk about this in the website team meeting- this is exactly one of the use cases for the list component.

mansona:
@ghislaineguerin I'm going to assign this to you. We have discussed the need for more flexibility with the card "component" i.e. card image, card detail etc.

Ideally we should be able to implement this page without the need for custom CSS in the website page. Let me know if you have any questions! 😂

mansona:
This PR is now ready to be reviewed and merged 🎉 

deploy-preview-375--ember-website netlify com_mascots_ (1)

I have avoided the need for "page CSS" by moving the `<TomsterList />` into a component.

Issue #313, opened 2019-04-18
Use unique titles for each page


j-brennan:
As a follow up to https://github.com/ember-learn/ember-website/issues/93, just opening this up to see if there is a desire to make each page have it's own dynamic page title?

Currently each page uses the title: `"Ember.js - A framework for ambitious web developers"`. Would it be useful to see different titles for each page. e.g. the Legal page uses `"Ember.js - Legal"` etc.

mansona:
@MelSumner I assume this is really a question for the SEO team? I can definitely work on a mechanism that would allow us to change the title dynamically (with a fallback) but I think we need to answer the question "do we need this" first 🤔

MelSumner:
@mansona we absolutely need this. It's an accessibility thing that we don't update the page titles. 

However, the page title format should be PageName - SiteName.

ryanolsonx:
Should we also make this change for the new rfc branch? I submitted a PR for master.

ryanolsonx:
Can someone close this (now that the PR has been merged)?

MelSumner:
@ryanolsonx yup! thank you for your help!

PR #379, opened 2019-08-12
Use unique titles for each page

ryanolsonx:
Adds titles (and subtitles) to each of the pages.  For example, Legal would be "Legal - Ember.js" and LTS releases would be "LTS - Releases - Ember.js".

I preserved the title for the home page "Ember.js - A framework for ambitious web developers".

See: #313

Side note: some Ember.js websites outside of this project use "Ember.js - Foo" while this PR uses "Foo - Ember.js".  We may need to change the naming pattern in the other places OR here.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <title>Ember.js - A framework for ambitious web developers</title>
     <meta name="description" content="Ember.js helps developers be more productive out of the box. Designed with developer ergonomics in mind, its friendly APIs help you get your job done—fast.">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta property="st:title" content="Ember.js: A framework for ambitious web developers" />
--- a/app/templates/about/legal.hbs
+++ b/app/templates/about/legal.hbs
@@ -1,3 +1,5 @@
+{{title "Legal"}}
+
 <h1 id="legal">Legal</h1>
 
 <section aria-labelledby="legal">
--- a/app/templates/application.hbs
+++ b/app/templates/application.hbs
@@ -1,3 +1,6 @@
+{{head-layout}}
+{{title "Ember.js"}}
+
 {{#es-header}}
   {{#es-navbar links=links home="/"}}
     {{!-- TODO: add search implementation --}}
--- a/app/templates/community.hbs
+++ b/app/templates/community.hbs
@@ -1,1 +1,3 @@
+{{title "Community"}}
+
 {{outlet}}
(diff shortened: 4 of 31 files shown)


ryanolsonx:
I know there's a major redesign happening currently, but I think this can still be of value for the current site. @wifelette @mansona

mansona:
@ryanolsonx don't worry too much about the redesign, if we get this merged into master then we will make sure that we either do a rebase or a merge into the new branch so it doesn't get lost 👍 

I have 2 concerns about this PR though 🤔 firstly looking at the titles they are all very "functional", and this is the reason why in the main issue I asked if the new SEO team would be looking at it (I'm not sure if this team ever got together in the end). The issue being is that the title is a very strong indication to the SEO, and we should be thinking about putting things in there that are what we want people to find Ember for search terms.

For example the new title for https://emberjs.com/releases/release/  is `Stable - Releases - Ember.js. Maybe that's what we actually want if we expect people to search for something like "Ember Stable" 🤔 but I'm wondering if there can be more thought put into it.

On the other side of this, considering this from a technical perspective instead of the content of each of the titles, I'm not sure if I want us to use ember-page-title for this 🤔 we have been using https://github.com/shipshapecode/ember-meta for other Ember sites which has allowed us to update things other than the title, including the Open Graph tags like `<meta property="st:title"`. The only issue is that it is more route and model based but not all of our routes have models 🤔 @rwwagner90 do you have any insights on how we might use ember-meta without routes or models? 

The reason why I would be interested in looking at ember-meta is that in this PR the `<title>` tag does not match the Open Graph metadata: 

```html
<meta property="st:title" content="Ember.js: A framework for ambitious web developers" />
<title>Stable - Releases - Ember.js</title>
```

Sorry if this is a bit of a brain dump but I've been thinking about your PR for a long time since you submitted it 😂

rwwagner90:
@mansona I added a `this.metaInfo` option to ember-meta, so you can set the meta separately from the model. I believe that should work for our use case.

ryanolsonx:
@mansona Thanks for the thorough insights on this PR.  I'll wait until the new SEO team gives their thoughts.  Once they respond, I can refactor this to ember-meta.

👍

rwwagner90:
@mansona is there an SEO team?

I think the general titles proposed here are good. Most sites do `Specific To Page - Website Name` formatted titles.

ryanolsonx:
@rwwagner90 Most of the titles should be in "Specific to Page - Website Name" format.  There are a few that are subpages which would get "Sub Page - Parent Page - Website Name".  This could change to a simple "Specific to Page - Website Name" everywhere.

In the example given above, "Stable - Releases - Ember.js" can become "Stable Release - Ember.js" or something.

rwwagner90:
@ryanolsonx I don't think changing is necessary. `Child - Parent - Website` is pretty typical. I think this would really help our SEO a ton. I am in favor 👍

ryanolsonx:
> `Child - Parent - Website` is pretty typical. I think this would really help our SEO a ton.

Not only would it help SEO, but it would help if a user bookmarks the page to have something that gives context.

PR #399, opened 2019-09-10
Update payment.hbs

MelSumner:
The form was not embedded correctly and the form doesn't work as a result. Trying to fix this.


wifelette:
For reference, here's what the form _should_ look like once it's sorted (in terms of fields, not design): https://tilde.wufoo.com/forms/tomster-and-zoey-commission-payments

MelSumner:
We'll need to update the styles but I just tested it and it works.

PR #403, opened 2019-09-12
Team page

ghislaineguerin:
This PR is for applying the styguide to the team page.


mansona:
@pichfl I know you have a new redesign for this page but I'm probably going to merge this as it's a major improvement to the broken design on the branch currently 👍 we can implement your new design as a follow up improvement PR

PR #407, opened 2019-09-24
Add codesandbox to the home page

NullVoxPopuli:
I think it'd be neat to have something interactive people could play with right away if they wanted.

I think this also encourages us to put some more effort into the `<WelcomePage />` :)

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -133,6 +133,22 @@
     <div class="col-3-large text-center">
       <img src="/images/ember-observer.svg" alt="ember observer">
     </div>
+
+    <div class="col-3-large">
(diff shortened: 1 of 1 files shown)


MelSumner:
What's the perf cost of this choice?

NullVoxPopuli:
codesandbox lazy-loads, so it's better than just loading the full sandbox.  I think the only way we could improve it is to use a lazy-loaded image placeholder (or a skeleton dom thing) and then swap out for the iframe on click or something. 
More than happy to make a component that does that, if that's desired

MelSumner:
I think we could put a prominent link that sends someone to codesandbox but I am not okay with it on the home page. + @jenweber we should review this with the rest of the learning team before moving forward.

PR #408, opened 2019-09-24
Adds ECG On-Demand to users

Robindfuller:
ECG ON-DEMAND® provide Immediate 12-lead ECG Interpretation and Same-Day Holter Monitor Analysis. We use Ember.js to create our in-house ECG analysis and workflow management software.


amyrlam:
@Robindfuller thanks for the submission! possible to amend the commit to kick off Percy? Not sure what happened. 

If not I can pull it down locally later!

chancancode:
Please remove the lock file changes

Robindfuller:
That should be done now

PR #409, opened 2019-09-25
Update guides links

chancancode:
Previously the guides link was pointing to the "Templating" section. Also added link to tutorial.


chancancode:
I think Percy is not going to run because we ran out of snapshots

PR #410, opened 2019-09-29
Remove Columbus meetup

kpfefferle:
We have decided to sunset the Columbus, OH meetup group. Nobody stepped up as organizer and it has been deleted from Meetup.com.


amyrlam:
thanks for keeping it updated! Seems like Percy is borked, but will merge

PR #423, opened 2019-10-07
Copy fix to /ember-users page

amyrlam:
The break was adding an unintentional space to the hyperlink.

WAS:
image

IS: 
From Percy:
image


amyrlam:
thanks @jenweber!

PR #424, opened 2019-10-07
Add BetterUp to users

amyrlam:
@bryanhickerson I think we'll need a larger image. 

Am wondering if you want to fill out the Airtable and a blurb for `This is where you fill in how you use Ember`.

Will send you a message on Slack


amyrlam:
will open a new PR

PR #449, opened 2019-10-25
Update guidelines page markup

sdebarros:
fixes https://github.com/ember-learn/ember-website/issues/437

Some comments:
- I didn't find a class to highlight the `General Goals` text as on the old page
- I'm not convinced about the padding between the header and the `<h1>` (seems a bit big), but it's due to the container, applying `padding-top-medium` didn't change anything.

### DIFF
--- a/app/templates/guidelines.hbs
+++ b/app/templates/guidelines.hbs
@@ -1,108 +1,114 @@
-<h1>Ember Community Guidelines</h1>
+<div class="container">
+  <h1>Ember Community Guidelines</h1>
+  <section class="margin-top-medium">
+    <p>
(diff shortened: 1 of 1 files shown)


MelSumner:
This seems to be missing a lot of markup.

sdebarros:
@MelSumner 👋 
Can you give me an idea of the markup that seems to be missing?
I did do minimal changes, as I couldn't find a page that I could use as an example of how this one should look. So I just added the container to center the content, and added sections. I'm happy to adjust it as needed.

mansona:
@MelSumner let me know if you have any concrete suggestions for improvement and I'll make sure to create a follow-up issue to ensure they are tackled 💪

PR #450, opened 2019-10-26
Update about/legal to comply with styleguide

### LINKED ISSUE #440: [redesign] About Legal Page - Update Markup
> If this is your first time tackling an issue that is part of the redesign project please read all the **Notes** section at the bottom of this issue before claiming it and starting to work on it

## To Do: 
- [ ] Update the markup for the `/about/legal` page

--- 

### Notes (important) 

This issue is part of the [Ember Website Redesign RFC](https://github.com/emberjs/rfcs/blob/master/text/0425-website-redesign.md) and is one of the **last remaining issues** before we are ready to release. As we are moving very quickly on this project you should comment on this issue that you are taking it on so that we do not have multiple people working on the same issue. **Note:** please don't take on an issue if you do not think you can complete it within 24 hours 🙏 

This issue is **strictly to only update the markup** of a page on the website and should be possible to do using only components and class helpers in the new [ember-styleguide](https://deploy-preview-145--ember-styleguide.netlify.com/).

If you find that you need to create page-specific styles, complete the page as best you can and open a PR **without** adding any styles. Make a note in your PR what was missing from the styleguide that prevented you from completing the page.

When creating your PR you should target the `redesign-rfc` branch. This way when your PR is merged it will not update `master` it will instead update the long-lived branch that is tracking the website redesign work.

jrjohnson:
@mansona is this the markup change you're looking for?

Biggest question I had was on the nesting of the `h1` tag. My general policy is to nest them within the section so that they label that area, but I wasn't able to find this in the styleguid.

Also unsure if leaving the `<strong>` was the right choice, or if it should replaced with a `<span class="bold">`.

Fixes #440

### DIFF
--- a/app/templates/about/legal.hbs
+++ b/app/templates/about/legal.hbs
@@ -1,6 +1,5 @@
-<h1 id="legal">Legal</h1>
-
-<section aria-labelledby="legal">
+<section class="container" aria-labelledby="legal">
+  <h1 id="legal">Legal</h1>
   <p>
     Ember is a registered trademark owned by <a href="http://www.tilde.io">Tilde Inc</a>.
   </p>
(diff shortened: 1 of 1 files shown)


app/templates/about/legal.hbs
```diff
@@ -1,41 +1,39 @@
-<h1 id="legal">Legal</h1>
+<div class="bg-muted">
```
MelSumner: I don't think we should need this wrapping div

app/templates/about/legal.hbs
```diff
@@ -1,41 +1,39 @@
...
-  </p>
-  <p>
-    This license does <strong>not</strong> include any uses of the Tomster included in the Guides. Please reach out for clarification on any ambiguous images before using.
-  </p>
-</section>
+    <h2>Documentation Images</h2>
+    <p>
+      The graphs, diagrams and charts in the <a href="/guides">Ember Guides</a> have been released under the Creative Commons Attribution-NonCommercial-ShareAlike (CC BY-NC-SA) license.
+    </p>
+    <img alt="CC, By, Non Commercial, Share Alike" class="pull-right" src="/images/legal/license-cc_by_nc_sa.png">
```
MelSumner: this should be `alt="" role="presentation"` since the preceding paragraph describes the license.

jrjohnson: 👍 I also worked it out to wrap correctly now.

app/templates/about/legal.hbs
jrjohnson: It was necessary because of the `bg-muted`, without it the default white background shows through, but I think I now understand that those `bg-*` colors are for calling out content and not for the default page. I've removed it. Thanks.

app/templates/about/legal.hbs
```diff
@@ -9,29 +8,32 @@
...
     The most current Ember license details <a href="https://github.com/emberjs/ember.js/blob/master/LICENSE">can be found on GitHub</a>.
   </p>
 
   <h2>Documentation Images</h2>
-
-  <p>
-    The graphs, diagrams and charts in the <a href="/guides">Ember Guides</a> have been released under the Creative Commons Attribution-NonCommercial-ShareAlike (CC BY-NC-SA) license.
-  </p>
-  <img alt="CC, By, Non Commercial, Share Alike" class="pull-right" src="/images/legal/license-cc_by_nc_sa.png">
+  <div class="layout-grid">
```
jrjohnson: What I would have preferred here was more control over the way text wraps around and image, but I didn't see that in the style guide, so I pulled it off the right as it was in the original version.

mansona: I think this is a reasonable approach using the grid helpers 🤔 I also think it works out quite well at lower resolutions.

May I ask why you're using `col-4` and not `col-5` so that you will fill the full width? I don't think it's a blocker for merging this but I'm just curious why you made that choice

jrjohnson: I wasn't actually sure what each grid size did. I played with it a bit on different device widths and this seemed like the best choice. Is there a way to indicate that the image should take up one unit to the right and then text should fill all remaining space? `col-fr4`, `col-fr1` for example.

mansona: interesting... it seems like we don't document the difference between `col-3` an `col-3-large` 🙈 I'll create an issue now to fix that  documentation

right now we have a 6 column system so if you want it the text to "fill all the remaining space" then you would make it use `col-5`

we then fall back to a 4 column system on smaller screens so if you want it to still have one column for the image and the "rest" for the text then you can do `col-5-large col-3` which just means "use 5 columns on larger screens but 3 columns on smaller screens 👍

jrjohnson: Got it! Thanks. I added that here.

mansona: Perfect that looks great! 🎉

PR #451, opened 2019-10-28
update styleguide to use new pre-release from npm

mansona:
This removes the need to use github urls and branch markers 🎉 and hopefully it should make Netlify a bit happier 😂

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "ember-qunit": "^4.4.1",
     "ember-resolver": "^5.0.1",
     "ember-source": "~3.10.0",
-    "ember-styleguide": "github:ember-learn/ember-styleguide#website-redesign-rfc",
+    "ember-styleguide": "^4.0.0-0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -60,7 +60,7 @@
     "ember-qunit": "^4.4.1",
     "ember-resolver": "^5.0.1",
     "ember-source": "~3.10.0",
-    "ember-styleguide": "github:ember-learn/ember-styleguide#website-redesign-rfc",
+    "ember-styleguide": "^4.0.0-0",
```
MelSumner: is this stable enough to do this?

mansona: Yes, this is a "pre-release" so it's not the "4.0" version yet. The next release will be `4.0.0-1` and so on until we are ready to release "v4.0.0" 👍 

You can see this when using `np` to release but also see more information here: https://semver.org/spec/v2.0.0.html#spec-item-9

PR #455, opened 2019-10-30
Update markup for Security page

mswanson:
fixes #442

I wasn't sure if there was a mockup somewhere of what this was supposed to look like. I just followed the class usage in #449 as an example. I'm happy to update it if necessary.

PR #458, opened 2019-10-31
Updated markup for About - Who Uses Ember page

ijlee2:
This PR addresses the issue #436 . In addition to the styleguide doc, I looked at PRs #450 and #455 for reference.

There are a few remaining issues that I couldn't solve on first attempt:

- The doc suggests, in order to use `card-link` option to make the entire `<EsCard>` interactive, I'd need to yield an `<a>` tag. Since an anchor needs text (content), I temporarily added the company's name. I'm not sure if there's an additional class name or attribute that I can pass to the card to make the text invisible.
  - An alternative approach may be to not use `@image` attribute, but instead add `` tag inside the `<a>` tag, as done previously?
- (1) I wasn't sure how to set class names to render fewer cards for smaller resolutions. (2) The current Ember website shows 5, 3, 2, or 1 card(s). With the new 6-column layout, I wasn't sure how many cards I was supposed to show and when.
- There is no bottom margin for `<ul class="list-unstyled layout-grid">`. As a result, whatever content would come next touches the cards too closely.

As an aside, making the entire `<EsCard>` interactive required me to surround the `<a>` tag with an `<h3>`. To avoid jumping from `h1` to `h3`, I added `<h2>` tag with the name `List of Companies`.

The text at the end of this page describes how one can add their company or project to the list. I wasn't sure if this text, semantically, belongs under `List of Companies`. For now, I placed it in a separate `h2`. Alternatively, I think we could look at using `<aside>` to place the text back under `List of Companies`?

[end of aside]

Before (desktop):
Screen Shot 2019-10-30 at 8 15 58 PM

Before (mobile, w=520px):
Screen Shot 2019-10-30 at 8 16 58 PM

Before (additional text at end):
Screen Shot 2019-10-30 at 8 17 29 PM

After (desktop):
Screen Shot 2019-10-30 at 8 16 11 PM

After (mobile, w=520px):
Screen Shot 2019-10-30 at 8 16 47 PM

After (additional text at end):
Screen Shot 2019-10-30 at 8 17 37 PM

### DIFF
--- a/app/templates/ember-users.hbs
+++ b/app/templates/ember-users.hbs
@@ -1,43 +1,41 @@
-<h1>
-  See Who's Using Ember.js
-</h1>
+<section class="container" aria-labelledby="ember-users">
+  <h1 id="ember-users">
+    See Who's Using Ember.js
+  </h1>
 
-<p>
-  Logos are added by company representatives. These companies may or may not be using Ember on their main web properties, but they're definitely using it somewhere in their organizations!
-</p>
-<p>
-  Feel free to track down folks who work at a specific company if you're looking for more detail on their specific usage.
-</p>
-
-<section aria-labelledby="Ember user list">
-  <ul class="user-showcase">
-    {{#each sortedModel as |user|}}
-      <li>
-        <a
-          class="user-showcase__link"
-          href={{user.url}}
-          rel="nofollow noopener"
-          target="_blank"
-        >
-          <img
-            src={{asset-map (concat "images/users/" user.image)}}
-            alt={{user.name}}
-          >
-        </a>
-      </li>
-    {{/each}}
-  </ul>
-
-  <p class="text-center">
-    To add your company or project to this page, <a href="https://github.com/ember-learn/ember-website" rel="nofollow">submit a Pull Request</a>.<br>
-    Be sure that your logo has a transparent background and includes adequate white space.<br>
-    Additional instructions are in the
-    <a
-      href="https://github.com/ember-learn/ember-website/blob/master/CONTRIBUTING.md#adding-a-user"
-      rel="nofollow noopener"
-      target="_blank"
-    >
-      contributing guide
-    </a>.
+  <p>
+    Logos are added by company representatives. These companies may or may not be using Ember on their main web properties, but they're definitely using it somewhere in their organizations!
   </p>
+  <p>
+    Feel free to track down folks who work at a specific company if you're looking for more detail on their specific usage.
+  </p>
+
+  <section aria-labelledby="section-ember-users-list-of-companies">
+    <h2 id="section-ember-users-list-of-companies">List of Companies</h2>
+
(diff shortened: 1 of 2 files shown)


app/templates/ember-users.hbs
```diff
@@ -1,43 +1,41 @@
...
-          class="user-showcase__link"
-          href={{user.url}}
-          rel="nofollow noopener"
-          target="_blank"
+    <ul class="list-unstyled layout-grid">
+      {{#each this.sortedModel as |user|}}
+        <EsCard
+          class="col-1-large"
+          @image={{asset-map (concat "images/users/" user.image)}}
+          @alt={{user.name}}
```
MelSumner: alt text should not be repeated as text - since the company name is here, the alt text should be blank and `role="presentation"` should be added.

ijlee2: Ah, ok. That makes sense. Will make that change soon.

On a related note, can I show the company's name in a card? The current Guides only shows the company's logo, so I wasn't sure of this change.

ijlee2: Note: Adding the presentation role will be completed in https://github.com/ember-learn/ember-styleguide/pull/243

MelSumner: > On a related note, can I show the company's name in a card? The current Guides only shows the company's logo, so I wasn't sure of this change.

I kind of like it, although I think we'd need a little more styling to make it work.

mansona: @ijlee2 for now we should go with your solution of using the yield to wrap the image only with a link instead of adding the company name.

This sort of change is not something that we can decide like this unfortunately and will require a large amount of data wrangling that is not in the scope of this issue/PR

ijlee2: @mansona Ok, sounds good. I'll try out `yield`ing when I get a chance today.

ijlee2: @mansona @MelSumner While yielding `<a>` and `` worked and matched better the current webpage, this presented a few problems in return.

I think these problems are minor (relative to my guess for this webpage's importance/current traffic) and could be addressed in a future ticket. What do you think?

- For a yielded `<a>` (without `h3` surrounding it), not *all* of the styles for `card-link` apply. I can see the background color change on hover, but I'm not able to click anywhere on the card to open the link. (I have to click on the company image.) This is because the pseudoelement styles that allow the card to be clicked anywhere are defined in selectors `h3 a::before` and `h3 a::after`.
- A yielded `` doesn't take full width, even with the parameter `full-image` on `<EsCard>`. Currently, that parameter and its corresponding CSS styling apply if we use `@image` attribute.
- When tabbing through the list, the blue box's position and dimension don't look great. It looks like the existing `outline` style doesn't apply well for a yielded `<a>` tag.
- I still haven't solved the problem of showing fewer cards across width for smaller resolutions.

Screen Shot 2019-11-02 at 9 07 02 AM

PR #461, opened 2019-11-02
Updated markup for About - Sponsors page

ijlee2:
This PR address the issue #443.

Screen Shot 2019-11-02 at 10 13 43 AM

### DIFF
--- a/app/templates/sponsors.hbs
+++ b/app/templates/sponsors.hbs
@@ -1,38 +1,57 @@
-<h1>Ember Sponsors and Friends</h1>
+<section class="container" aria-labelledby="ember-sponsors-and-friends">
+  <h1 id="ember-sponsors-and-friends">
+    Ember Sponsors and Friends
+  </h1>
 
-<p class="intro">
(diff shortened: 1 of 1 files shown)


app/templates/sponsors.hbs
```diff
@@ -1,38 +1,57 @@
...
+            </a>
+          </h3>
+          <p>{{sponsor.term}}</p>
+        </EsCard>
+      {{/each}}
+    </ul>
+  </section>
 
-  <h2 class="text-center">Past Sponsors</h2>
+  <section aria-labelledby="ember-sponsors-and-friends-past-sponsors">
```
ijlee2: The markup is the same as above, with past instead of current.

mansona:
@ijlee2 this was originally done against the wrong branch 🙈 not to worry! I used some git-fu and fixed it for you 🎉 

We also discussed it at our weekly meeting today and we decided we would get this merged and add some more tweaks to this later to improve the layout 👍 thanks again for your work 🎉

PR #467, opened 2019-11-12
Updated markup for Learn - Examples page

ijlee2:
This PR addresses the issue #439 . Here, I used `<EsCard>` component to represent the information.

If we want to make the webpage look more like what it is now, we might add a grid layout around each example, with 2 columns for the image and 4 columns for the text. However, I thought this might cause the text to be unreadable at mobile resolution, which is what happens currently. I have not tried this approach to see whether the issue would actually occur with the new styleguide.

Desktop (before):

Desktop (after):

Mobile (before):

Mobile (after):

### DIFF
--- a/app/templates/learn/examples.hbs
+++ b/app/templates/learn/examples.hbs
@@ -1,34 +1,45 @@
-<h1 id="examples">Examples</h1>
+<section class="container" aria-labelledby="learning-emberjs-examples">
+  <h1 id="learning-emberjs-examples">Examples</h1>
 
-<p>In this section you will find applications that are maintained by the Ember.js teams with the help of contributors. While software is always a work in progress, the goal is to showcase patterns and solutions applied in real-world applications.</p>
-<p>Whether you're simply interested in checking out how some feature is implemented, or you're looking to contribute, one of these projects might pique your interest!</p>
+  <p>In this section, you will find applications that are maintained by the Ember.js teams with the help of contributors. While software is always a work in progress, the goal is to showcase patterns and solutions applied in real-world applications.</p>
 
-<div class="showcase">
-  {{#each model as |showcase|}}
-    <div class="showcase--item">
-      <div class="showcase--screenshot">
-        <img
-          class="showcase--image"
-          src={{asset-map (concat "images/showcase/" showcase.image.src)}}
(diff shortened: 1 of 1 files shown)


ijlee2:
As an aside, I don't think there is a way to land on this webpage through `<LinkTo>` anymore. Should we consider removing this route? There's duplicate information on `/learn` route.

app/templates/learn/examples.hbs
```diff
@@ -1,34 +1,45 @@
...
-      </div>
-      <div class="showcase--description">
-        <h3>
-          <a href={{showcase.demo}}>{{showcase.name}}<i class="icon-link-ext"></i></a>
-        </h3>
-
-        <p>Repository: <a href={{showcase.repository}}>{{showcase.repository}}</a></p>
-
-        <p>
-          {{showcase.description}}
```
ijlee2: The `showcase` model doesn't have an attribute named `description`.

MelSumner:
> As an aside, I don't think there is a way to land on this webpage through `<LinkTo>` anymore. Should we consider removing this route? There's duplicate information on `/learn` route, which _can_ be accessed through the top navigation bar.

@ijlee2 can you explain more? Discord is also fine.

ijlee2:
> > As an aside, I don't think there is a way to land on this webpage through `<LinkTo>` anymore. Should we consider removing this route? There's duplicate information on `/learn` route, which _can_ be accessed through the top navigation bar.
> 
> @ijlee2 can you explain more? Discord is also fine.

@MelSumner Sorry for a very late reply!

If I remember right, I wasn't able to find a way to reach the Examples page, starting from emberjs.com homepage. If this is the case, I was thinking that a user would need to know the URL to land on the page, or land via Google search.

I'm not too familiar with the entire site, so it's possible that I missed the way to reach the page.

locks:
> There's duplicate information on /learn route, which can be accessed through the top navigation bar.

The goal was to move the showcase to its own page and link to it from the learn page.

ijlee2:
@locks Ah, ok. Makes sense, thanks!

mansona:
Closing in favour of https://github.com/ember-learn/ember-website/pull/499

PR #491, opened 2019-12-22
Redirect /team to /teams

### LINKED ISSUE #487: Move `/team` to `/teams` and redirect `/team`?
I keep (stupidly) going to `emberjs.com/teams` and realizing that I have to use `/team` instead. We have had many teams for quite a while now, so I think it might be time to move `/team` to `/teams` with a redirect. 

Thoughts?

locks:
Fixes #487.

### DIFF
--- a/.netlifyredirects
+++ b/.netlifyredirects
@@ -10,6 +10,7 @@
 /zoey       /mascots
 /brand      /logos/
 /branding   /logos
+/team       /teams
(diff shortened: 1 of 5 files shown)


jenweber:
looks like there is some work to do to make the tests happy. Perhaps rename the route file as well?

locks:
@jenweber can you give it another look?

PR #496, opened 2020-01-05
remove ember-cli-ifa

mansona:
This is required to prevent the issues where the JS was crashing because of integrity checks. If you want more information on why this is required you can check out this issue: https://github.com/ef4/prember/issues/52


mansona:
I'm going to merge this myself if Percy passes as it is a significant issue that is breaking the behaviour of website currently 🙈

Issue #444, opened 2019-10-24
[redesign] Mascots Commission Pages - Update Markup


mansona:
> If this is your first time tackling an issue that is part of the redesign project please read all the **Notes** section at the bottom of this issue before claiming it and starting to work on it

## To Do: 
- [x] Update the markup for the 'mascots/commission-sent' page
- [x] Update the markup for the 'mascots/commission'  page
- [x] Update the markup for the 'mascots/faq'  page
- [x] Update the markup for the 'mascots/payment-sent'  page
- [x] Update the markup for the 'mascots/payment'  page
--- 

### Notes (important) 

This issue is part of the [Ember Website Redesign RFC](https://github.com/emberjs/rfcs/blob/master/text/0425-website-redesign.md) and is one of the **last remaining issues** before we are ready to release. As we are moving very quickly on this project you should comment on this issue that you are taking it on so that we do not have multiple people working on the same issue. **Note:** please don't take on an issue if you do not think you can complete it within 24 hours 🙏 

This issue is **strictly to only update the markup** of a page on the website and should be possible to do using only components and class helpers in the new [ember-styleguide](https://deploy-preview-145--ember-styleguide.netlify.com/).

If you find that you need to create page-specific styles, complete the page as best you can and open a PR **without** adding any styles. Make a note in your PR what was missing from the styleguide that prevented you from completing the page.

When creating your PR you should target the `redesign-rfc` branch. This way when your PR is merged it will not update `master` it will instead update the long-lived branch that is tracking the website redesign work.

mansona:
Fixed in #506

PR #506, opened 2020-01-09
Mascots Commission Pages

mansona:
Fixes https://github.com/ember-learn/ember-website/issues/444

This PR really falls into the "not great, but good enough for now" category 🙈 It is no worse than current production, and actually fixes a bug on https://emberjs.com/mascots/payment/ 😂

### DIFF
--- a/app/styles/legacy/form.css
+++ b/app/styles/legacy/form.css
@@ -1,0 +1,133 @@
+
+.tomster-form {
+  margin-top: 1.5em
+}
+
+@media screen and (min-width:0px) and (max-width:991px) {
+  .tomster-form {
+    padding: .5em
+  }
+
+  .tomster-form label, .tomster-form legend {
+    text-align: left
+  }
+}
+
+.tomster-form ul {
+  padding: 0;
+  margin: 0 !important;
+  list-style: none
+}
+
+.tomster-form label, .tomster-form legend {
+  margin-top: 1.5em;
+  margin-bottom: .4em;
+  display: block;
+  font-weight: 700
+}
+
+.tomster-form label.error, .tomster-form legend.error {
+  color: #c7341f;
+  font-size: 12px;
+  margin-top: -.9em
+}
+
+.tomster-form label .req, .tomster-form legend .req {
+  color: #c7341f
+}
+
+.tomster-form input[type=radio] {
+  display: inline-block
+}
+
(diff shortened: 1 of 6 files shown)


app/styles/legacy/form.css
```diff
@@ -0,0 +1,128 @@
...
+
+.tomster-form .inline-field input[type=text] {
+  width: auto
+}
+
+.tomster-form .inline-field label {
+  line-height: 1.6em
+}
+
+.hidden{display:none}
```
MelSumner: we def shouldn't have hidden or clearfix in the website css. if we need these we need to put them in ember-styleguide, and then have a test that checks for duplicates.

app/templates/mascots/payment-sent.hbs
```diff
@@ -1,9 +1,6 @@
-<h1>Tomster and Zoey Commission Payments</h1>
-
-<div class="tomster-payment section">
-  <div class="left-col">
-    <p>Thanks much! Looking forward to getting to work :)</p>
-    <p><a href="/mascots">&larr; Back to the other Tomsters and Zoeys</a></p>
-  </div>
-</div>
+<section class="container">
+  <h1>Tomster and Zoey Commission Payments</h1>
```
MelSumner:
```suggestion
  <h1 id="commision-payments">Tomster and Zoey Commission Payments</h1>
```

app/templates/mascots/payment-sent.hbs
```diff
@@ -1,9 +1,6 @@
-<h1>Tomster and Zoey Commission Payments</h1>
-
-<div class="tomster-payment section">
-  <div class="left-col">
-    <p>Thanks much! Looking forward to getting to work :)</p>
-    <p><a href="/mascots">&larr; Back to the other Tomsters and Zoeys</a></p>
-  </div>
-</div>
+<section class="container">
```
MelSumner:
```suggestion
<section class="container" aria-labelledby="commision-payments">
```

app/styles/legacy/form.css
mansona: I've updated these to a) make sure they are scoped to only these wufoo forms and b) that they have a comment making sure it's clear that they are legacy and need to be deleted later

Essentially we need to properly sit down and design these pages, and there should **never** be a clearfix class helper in the website or the styleguide if we're doing things right. Also this `.hidden` class seems to be some sort of security feature to protect a bit against bots 🤔

PR #507, opened 2020-01-10
Rename and redirect logos to branding

locks:


tests/acceptance/visual-routes-test.js
```diff
@@ -23,7 +23,7 @@ module('Acceptance | visual regression', function (hooks) {
     { title: '/index.html', route: '/'},
     { title: '/learn.html', route: '/learn/' },
     { title: '/legal.html', route: '/about/legal/' },
-    { title: '/logos.html', route: '/logos/' },
```
mansona: you should not change the left hand side of this, if you do then we lose continuity. The left hand side is just the title for Percy to identify the snapshot by name. You can see that the left hand side (title) is already out of sync with the route in a few other places 👍

tests/unit/routes/branding-test.js
```diff
@@ -1,11 +1,11 @@
 import { module, test } from 'qunit';
 import { setupTest } from 'ember-qunit';
 
-module('Unit | Route | logos', function(hooks) {
+module('Unit | Route | branding', function(hooks) {
```
mansona: we should delete this test because it's empty

locks: https://github.com/ember-learn/ember-website/pull/508 ;)

locks:
Seems fine to wait for the redesign :)

tests/acceptance/visual-routes-test.js
locks: that doesn't sound great tho :P

mansona: it doesn't matter if it sounds great, it is how it's supposed to work :) it has been this way since the Ruby->Ember migration 👍 please fix

MelSumner:
@mansona let's figure out how to make this easier- we shouldn't have to jump through so many hoops on an Ember app when this is a simple server-side redirect in pretty much everything else. Even identifying that this is a point of friction in Ember apps in general would be useful because we can file an issue and get this worked on.

mansona:
I don't think this is a function of making it difficult in "Ember Apps", I think this is a challenge that you would face in any app that is both Server-Side Generated and has client-side rendering. We just need to make sure that both things are in sync.

The main friction in my mind is that we need to define the value in the change before taking the time (and being careful) to actually make the change. We recently had a similar issue with https://github.com/ember-learn/ember-website/pull/491/files which ended up working in the end but I seem to remember it causing a bug 🤔 (can't find the link to the bug or the bugfix now) 

In that case the justification wasn't that strong, but there was at least **some** justification for the change! And the outcome is better because both `/team` and `/teams` work now 👍

PR #508, opened 2020-01-10
Remove boilerplate files

locks:


locks:
@amyrlam I was doing some other changes and noticed that there are a lot of boilerplate files in the project, mainly auto-generated tests and controller files. I think this is cleaner for everyone!

amyrlam:
@locks cool sg! thanks for the explanation!

PR #509, opened 2020-01-10
update ember-styleguide

mansona:


mansona:
this now depends on https://github.com/ember-learn/ember-styleguide/pull/262 before it can be merged 👍

Issue #520, opened 2020-01-16
Create "Meet Addons" tab component for landing page


mansona:
![image (2)](https://user-images.githubusercontent.com/594890/72536594-6e872800-3872-11ea-8118-d911f52fab27.png)

mansona:
Fixed in #530

PR #530, opened 2020-01-19
Addon tabs component

mansona:
This makes the "Meet Addons" section image a real component with a tabbed UI and no longer just an image

Fixes https://github.com/ember-learn/ember-website/issues/520


app/components/addon-tabs.js
```diff
@@ -0,0 +1,6 @@
+import Component from '@ember/component';
```
MelSumner: This should definitely be in the style guide addon.

mansona:
@MelSumner I disagree that this should be in the styleguide 🙈 We can discuss it more on Thursday but I'll summarise it quickly here too.

Since the [last time that I tried to put this in the styleguide](https://github.com/ember-learn/ember-styleguide/pull/157), I have been thinking a lot about this implementation and have even tried it out with a client's app. There is a lot to be desired when coming up with a truly generic implementation of Tabs that could work like this does and could be adapted for multiple places and my experimentation has not lead me to find something that is good enough to be universal yet.

The implementation in this PR is not really generic enough to move into the styleguide and definitely doesn't have a good API for people to use. I would be very much against moving this to the styleguide with a promise of "other use cases for tabs" without first having an example of one of those use cases and building something that works for both cases. If we did it now I feel very strongly that it would be a big premature optimisation 🙈

pichfl:
I also see this more in the website repository than in the style guide for now. We can easily extract them from here and move them to the style guide later on.

MelSumner:
We want this component for other websites, which is why I think it should be in the styleguide.

PR #534, opened 2020-01-21
Remove builds app from showcase as it has been archived and merged with the website

locks:


MelSumner:
@mansona this doesn't work because it's still the old page. @locks we can do this for both sites, one min.

mansona:
I'm pretty sure that if we merge https://github.com/ember-learn/ember-website/pull/535 it will cause a conflict between the redesign branch and master and we would need to resolve it. I **think** the best way to make sure there is no conflict is to cherry-pick instead of merging a new commit 🤔 

@MelSumner what do you mean by "still the old page"?

PR #535, opened 2020-01-23
Delete builds.md

MelSumner:
removing builds.md


mansona:
@locks I still think this is going to cause a conflict and we should do a merge from master into `redesign-rfc` before merging again anyway.

I really think we should **not** merge this

locks:
Let's do that, then!

PR #555, opened 2020-02-09
improve homepage button spacing on very small screens

nickschot:
On very small screen (iphone SE specifically) the button groups on the homepage would wrap with some spacing issues.

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -5,8 +5,8 @@
   @detail="Ember.js is a productive, battle-tested JavaScript framework for building modern web applications. It includes everything you need to build rich UIs that work on any device."
   class="bg-dark"
 >
-  <a href="https://guides.emberjs.com/release/getting-started/quick-start/" class="es-button">Quickstart</a>
-  <a href="https://guides.emberjs.com/release/tutorial/ember-cli/" class="es-button-secondary ml-1">Read the Tutorial</a>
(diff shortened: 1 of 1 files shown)


app/templates/index.hbs
```diff
@@ -159,7 +159,7 @@
 <section>
   <div class="layout container">
     <div class="lg:col-4 lg:start-2 text-center">
-      <a href="https://emberconf.com" class="es-button">
+      <a href="https://emberconf.com" class="es-button mb-1">
```
mansona: Shouldn't this use the same strategy as above? the first one has mb-1 and mr-1 instead of the second one having ml-1? 🤔

nickschot: These buttons we're somewhat annoying to get right with just the functional margin classes as they are centered on screen (the others are left aligned). I chose to give the smaller button the margin. It will appear slightly off center but this is the best we can do without custom css or a more "elaborate" flexbox setup.

Screenshot 2020-02-12 at 12 12 29

nickschot: Just to clarify: with the mr-1 approach the bigger button would be off center which would be more noticable.

mansona: so the only way to fix this perfectly would be with the "button group" implementation that we discussed previously 🤔 hmm

I think I'm going to merge this and create an issue to continue this after the release 👍

mansona: https://github.com/ember-learn/ember-styleguide/issues/283

PR #571, opened 2020-02-17
Update "books and blogs" content on home page

MelSumner:
Adds more unofficial resources to the "Books and blogs" list.

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -222,7 +222,9 @@
         <h4 class="text-md">Books & blogs</h4>
 
         <ul class="list-unstyled">
-          <li><a href="https://www.balinterdi.com/rock-and-roll-with-emberjs/">Rock and Roll with Ember.js</a></li>
+          <li><a href="https://www.balinterdi.com/rock-and-roll-with-emberjs/" title="a book that helps you learn Ember">Rock and Roll with Ember.js</a></li>
+          <li><a href="http://emberatlas.com/" title="unofficial resource that provides additional information for developers">Ember Atlas</a></li>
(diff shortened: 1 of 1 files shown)


app/templates/index.hbs
```diff
@@ -222,7 +222,9 @@
         <h4 class="text-md">Books & blogs</h4>
 
         <ul class="list-unstyled">
-          <li><a href="https://www.balinterdi.com/rock-and-roll-with-emberjs/">Rock and Roll with Ember.js</a></li>
+          <li><a href="https://www.balinterdi.com/rock-and-roll-with-emberjs/" title="a book that helps you learn Ember">Rock and Roll with Ember.js</a></li>
+          <li><a href="http://emberatlas.com/" title="unofficial resource that provides additional information for developers">Ember Atlas</a></li>
```
mansona: Just leaving my comment on this the same as the previous PR for this content: I think we need to get consensus from the @ember-learn/learning-core-team before we merge this

MelSumner: @mansona agree- thought we already had it though. Thanks for checking! 

@jenweber can you weigh in here please, since you do the work on the guides?

jenweber: My own opinion is that overall it's fine to show Atlas in a list of community resources, since that's what it is. I'd like to see that clarified in the description on the project landing page.

However, right now, the information there is quite sparse so I'd also recommend listing something else for now. I count only 6ish articles at the moment. More than half of the topics have no content.

MelSumner:
```suggestion
```

PR #572, opened 2020-02-17
Update addon-tabs.css

MelSumner:
Ensures that tab content is vertically centered

### DIFF
--- a/app/styles/components/addon-tabs.css
+++ b/app/styles/components/addon-tabs.css
@@ -12,6 +12,7 @@
 }
 
 .addon-tabs--tab {
+  align-content: center;
   background: none;
   border-color: #54565E;
   border-width: 1px;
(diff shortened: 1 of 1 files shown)


mansona:
What is this trying to fix? 🤔 I don't see a diff in Percy and I can't see any difference when I run the Netlify preview and disable this rule

MelSumner:
Firefox

app/styles/components/addon-tabs.css
```diff
@@ -12,6 +12,7 @@
 }
 
 .addon-tabs--tab {
+  align-content: center;
```
mansona:
```suggestion
  align-items: center;
```

With `align-content` it looks like this: 

Screenshot 2020-02-17 at 16 49 04

I assume that is no-change from what it looked like before? 

With `align-items` it looks like this: 

Screenshot 2020-02-17 at 16 49 42

Which I assume fixes the issue?

PR #575, opened 2020-02-17
Make the video whitespace consistent

Windvis:
I noticed the [video on the community page](https://deploy-preview-361--ember-website.netlify.com/community) had inconsistent whitespace which I assume is a remnant from the previous version.

I also noticed the images are a bit blurry do to them being upscaled and that especially noticable on mobile devices. Are there any plans of them being replaced with svg versions?

If not, I personally wouldn't upscale them and just center the image on mobile, and left / right align them on larger devices (depending on how they are positioned next to the text). Thoughts? @MelSumner @mansona 

PS. I know things like this is very nitpicky so let me know if it shouldn't be reported :smile:

### DIFF
--- a/app/templates/community/index.hbs
+++ b/app/templates/community/index.hbs
@@ -110,7 +110,7 @@
 </div>
 
 <div class="container">
-  <div class="embed-video mt-5">
+  <div class="embed-video">
     <iframe width="854" height="480" src="https://www.youtube.com/embed/rY5D38RQoEg?rel=0" frameborder="0" allowfullscreen></iframe>
   </div>
 </div>


mansona:
@Windvis it's good that you're noticing finer details, but I would suggest just creating an issue in `ember-website` and not trying to fix it before release. I am pretty sure that the upscaling that you're talking about is also present on the current production website and we're trying to stay 100% focused on blockers for the release of the redesign right now 👍 

Is there any reason why this PR is marked as WIP? It looks good and should be merged but I don't want to do it before it's ready

Windvis:
I wasn't sure if I needed to add the image changes to this MR. :smile:

mansona:
I prefer to keep PRs nice and focused whenever possible so I would prefer those image improvements in a different PR 🎉 

But like I said, there are other things way more important than those image improvements that are actively blocking launch

PR #578, opened 2020-02-17
added cta-banner component

MelSumner:
adds the CTA for the survey.

### DIFF
--- a/app/components/cta-banner.js
+++ b/app/components/cta-banner.js
@@ -1,0 +1,4 @@
+import Component from '@ember/component';
+
+export default Component.extend({
+});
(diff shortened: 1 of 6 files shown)


mansona:
@MelSumner it looks like we already have an `ember-survey` component that adds functionality for "number of days left" for the survey https://github.com/ember-learn/ember-website/blob/redesign-rfc/app/components/ember-survey.js

This also seems to be implemented in a way that only requires config to be added to enable it 🤔 do you think that we could make use of this existing component instead of creating a new one?

MelSumner:
I'm going to close this issue, @mansona is going to work on a version that combines some things we already have.

PR #579, opened 2020-02-17
swapped out pngs for svgs

MelSumner:
swaps out PNGs for updated SVG files.


mansona:
I've approved it but I think @pichfl mentioned that he might want to take a look before we merge 🤔

PR #580, opened 2020-02-17
added cta for EmberConf

MelSumner:
adds CTA for EmberConf


app/styles/components/cta-emberconf.css
```diff
@@ -0,0 +1,197 @@
...
+  .emberconf-announcement .emberconf-date {
+      flex-basis:auto;
+      margin-left: 1.25rem;
+      margin-top: 0;
+      text-align: left;
+  }
+}
+
+.emberconf-announcement .emberconf-date .line-1 {
+  color: #9b2918;
```
mansona:
```suggestion
  color: var(--color-brand-hc-dark);
```
Again I don't really know about this one but I wonder if we should try to make it match more the other "orange" colours in the section (the link and the EmberConf logo).

app/styles/components/cta-emberconf.css
```diff
@@ -0,0 +1,197 @@
...
+  flex-basis: 100%;
+  margin-bottom: 0;
+  margin-top: 0
+}
+
+.emberconf-announcement .announcement-description p+p {
+  margin-top: .5rem
+}
+
+/* TODO I really shouldn't have to give this much specificity */
```
mansona: I think your bg-none helper in the styleguide should be made work with this 🤔 we already had to tweak the specificity of some of the helpers so that they did what you expect.

If you wanted to create this as an issue I would be happy to tackle this **after** we release 👍

app/styles/components/cta-emberconf.css
```diff
@@ -0,0 +1,197 @@
+.emberconf-announcement {
+  align-items: center;
+  background: var(--color-gray-300);
```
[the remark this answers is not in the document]
MelSumner: I didn't pick the next background because I didn't want someone to think that I just oops got it in the wrong place, and the same next color also felt too matchy-matchy instead of coordinated. Make sense?

app/styles/components/cta-emberconf.css
MelSumner: yeah, I was also surprised that it didn't. It's how the `a` is styled in general, scoped to main :(

app/templates/components/cta-emberconf.hbs
```diff
@@ -0,0 +1,20 @@
...
+      </div>
+    </div>
+    <div class="announcement-description">
+      <div class="text">
+        <p>EmberConf is back for another year of talks, training and fun!</p>
+        <p><a href="//emberconf.com">Register at EmberConf.com </a></p>
+      </div>
+    </div>
+  </div>
+</div>
```
MelSumner:
```suggestion
</div> 

```

app/styles/components/cta-emberconf.css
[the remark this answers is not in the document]
mansona: yea makes sense 👍

app/styles/components/cta-emberconf.css
mansona: issue filed https://github.com/ember-learn/ember-styleguide/issues/294

PR #583, opened 2020-02-18
run SVGO on new svgs

mansona:


mansona:
I was trying this to see if Travis and Percy would run but it seems like they don't 🤔

PR #585, opened 2020-02-18
AddonTabs keyboard improvements

### LINKED ISSUE #567: Make the AddonTabs component fully keyboard accessible
It seems that it's not possible to switch to the other tabs with keyboard based navigation alone at the moment. 

image

Expected behavior:
- Pressing `TAB` should cause focus to the row of tabs
- Left or right arrow keys should cause movement inside of the tab list 
- pressing `ENTER` should activate the new tab and show the corresponding content

The current behavior of pressing `TAB` while a tab is active is correct (it focuses on the tab content itself and then continues to travel down the tabindex. 

References:
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role
- https://www.w3.org/TR/wai-aria-practices/#tabpanel

Windvis:
Fixes #567

### DIFF
--- a/app/components/addon-tabs.js
+++ b/app/components/addon-tabs.js
@@ -1,6 +1,50 @@
 import Component from '@ember/component';
 
+const KEY = {
+  ARROW_LEFT: 'ArrowLeft',
+  ARROW_LEFT_IE11: 'Left',
+  ARROW_RIGHT: 'ArrowRight',
+  ARROW_RIGHT_IE11: 'Right'
+}
+
 export default Component.extend({
   currentTab: 0,
   tagName: '',
+
+  handleKeyboardNavigation (event) {
+    let { key } = event
+    let shouldHandleKey = Object.values(KEY).includes(key)
+
+    if (shouldHandleKey) {
+      let tabElement = event.target
+      let tabElements = tabElement.parentElement.children 
+      let totalTabs = tabElements.length
+
+      let newTab = this.currentTab
+      if (key === KEY.ARROW_LEFT || key === KEY.ARROW_LEFT_IE11) {
+        newTab -= 1
+      } else if (key === KEY.ARROW_RIGHT || key === KEY.ARROW_RIGHT_IE11) {
+        newTab += 1
+      }
+
+      newTab = wrapTabIndex(newTab, totalTabs)
+      this.set('currentTab', newTab)
+      tabElements[newTab].focus()
+    }
+  }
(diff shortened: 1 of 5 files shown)


app/components/addon-tabs.js
```diff
@@ -1,6 +1,48 @@
 import Component from '@ember/component';
 
+const KEY = {
+  ARROW_LEFT: 'ArrowLeft',
```
Windvis: Since IE11 support is required, this should be converted to KeyCodes instead

app/components/addon-tabs.js
```diff
@@ -1,6 +1,48 @@
...
+  ARROW_RIGHT: 'ArrowRight'
+}
+
 export default Component.extend({
   currentTab: 0,
   tagName: '',
+
+  handleKeyboardNavigation (event) {
+    let { code } = event
+    let shouldHandleKey = Object.values(KEY).includes(code)
```
Windvis: These need to be polyfilled for IE11 support

app/templates/components/addon-tabs/panel.hbs
```diff
@@ -1,9 +1,12 @@
...
-</div>
\ No newline at end of file
+{{#let (eq @currentTab @tabId) as |isTabActive|}}
+  <div
+    id="addon-panel-{{@tabId}}"
+    role="tabpanel"
+    tabindex="0"
+    aria-labelledby="addon-tab-{{@tabId}}"
+    class="addon-tabs--content {{if isTabActive "active-tab"}}"
+    hidden={{not isTabActive}}
```
Windvis: The hidden attribute seemed to be missing in the previous implementation so I added that as well.

Windvis: The `not` helper isn't actually coming from the project. It only works because an addon depends on `ember-truth-helpers` already.

app/components/addon-tabs.js
Windvis: Changed to use KeyboardEvent.key instead and added support for the IE11 version. Adding a polyfill might be a better solution though.

Windvis:
I'll install ember-test-selectors and add some tests for the component as well.

app/templates/components/addon-tabs/panel.hbs
Windvis: Opened #589 for this.

PR #591, opened 2020-02-20
Redesign/addon tabs styling improvements

Windvis:
I noticed some weird behaviour in Firefox when clicking the AddonTabs buttons. The text was shifting 1 pixel, and the borders changed color. The borders also had a different color in Chrome.


Windvis:
Hmm, it seems like this also updated ember-styleguide to v4.0.0-11 which causes the percy conflicts. Not sure how to fix the dependency problem without purging the lock file though. Should the style guide dependency be pinned to a specific version?

PR #593, opened 2020-02-21
install ember-cli-babel-polyfills

Windvis:
This makes the JS actually work in IE11 😄 

Did some production build tests:

## No polyfills:
```
File sizes:
 - dist/assets/vendor-fc2a4683c698bde849b7bcf627dd1c87.js: 925.21 KB (232.47 KB gzipped)
```

## With ember-cli-babel's `includePolyfills` option:
```
File sizes:
 - dist/assets/vendor-4bf71e5ab76e234f522bf501919304fe.js: 1010.38 KB (261.65 KB gzipped)
```

## With ember-cli-babel-polyfills
```
File sizes:
 - dist/assets/polyfill-evergreen-48158e9aad393fe17273e5052cc3c308.js: 87 B (103 B gzipped)
 - dist/assets/polyfill-legacy-a1c82203cebbb94c3526d222ca1645d9.js: 61.47 KB (19.7 KB gzipped)
 - dist/assets/polyfill-shared-741350dc959a1d1eabe81b2dbebad156.js: 48.27 KB (17.91 KB gzipped)
 - dist/assets/vendor-08b7ed85b4bc1a3e4a8ceb1cfa6e51ef.js: 922.66 KB (232.2 KB gzipped)
```
(polyfill-legacy will not be loaded in evergreen browsers)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,84 +5,56 @@
   "requires": true,
   "dependencies": {
     "@babel/code-frame": {
-      "version": "7.0.0",
-      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.0.0.tgz",
-      "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
+      "version": "7.8.3",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.8.3.tgz",
+      "integrity": "sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "^7.0.0"
+        "@babel/highlight": "^7.8.3"
       }
     },
     "@babel/compat-data": {
-      "version": "7.8.1",
-      "resolved": "https://registry.npmjs.org/@babel/compat-data/-/compat-data-7.8.1.tgz",
-      "integrity": "sha512-Z+6ZOXvyOWYxJ50BwxzdhRnRsGST8Y3jaZgxYig575lTjVSs3KtJnmESwZegg6e2Dn0td1eDhoWlp1wI4BTCPw==",
+      "version": "7.8.5",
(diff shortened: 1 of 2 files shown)


Windvis:
I also had to update the transitive dependencies because of some older @babel/core and @babel/preset-env versions causing build errors.

Windvis:
It looks like this also updated ember-styleguide to v4.0.0-11 which causes the percy conflicts. I'm not sure how to fix the babel dependency errors without purging the lock file though 🤔. Should the ember-styleguide dependency be pinned to a specific version to prevent accidental bumps?

MelSumner:
try putting this into ember-styleguide instead of into the website, see if that makes it better.

MelSumner:
@Windvis @mansona let's figure this one out so we can merge or close. 👍

mansona:
the asset size output: 

Files that got Bigger 🚨:

File | raw | gzip
--- | --- | ---
auto-import-fastboot.js|+7.69 kB|+2.18 kB
ember-website.js|+680 B|+3 B
ember-website-fastboot.js|+115 B|+38 B
polyfill-evergreen.js|+87 B|+103 B
polyfill-legacy.js|+62.9 kB|+20 kB
polyfill-shared.js|+49.4 kB|+18.3 kB
vendor.css|+22 B|+10 B

Files that got Smaller 🎉:

File | raw | gzip
--- | --- | ---
vendor.js|-2.61 kB|-265 B


Files that stayed the same size 🤷‍:

File | raw | gzip
--- | --- | ---
ember-website.css| 0 B| 0 B

mansona:
So this should go in the ember-website and **not** the ember-styleguide, so we shouldn't move it there 👍 

I am curious when trying to test this, what JS is not actually working right now? i.e. if I were to look at the deploy preview vs the live site what isn't working on production and is fixed with this?

I think this needs to be rebased on master now that we have merged and the PR should also target master too. After that is done I don't see any reason why this can't be merged 🤔

Windvis:
@mansona When I open the website in IE11 I see the following error in the console:
`Error while processing route: index Object doesn't support property or method 'endsWith' TypeError: Object doesn't support property or method 'endsWith'`. Which is why I assumed we needed to include polyfills for IE.

You're not seeing that error? Clicking the dropdowns / tabs works in IE?

I find it surprising that the `polyfill-shared.js` is so large, since everything seems to work in evergreen browsers. But that's just how that addon works I guess 😄. And it's better than including the legacy polyfills as well (which is what the ember-cli-babel includePolyfills option does).

I can redo this tomorrow if you want.

Windvis:
Redid this in #611

PR #607, opened 2020-02-26
adding lighthouse CI

mansona:


mansona:
If anyone wants to see the full report that is generated check out https://ember-lighthouse-ci-server.herokuapp.com/app/projects/ember-website/compare/81bedc7b-8409-44e7-8730-142a3cf96a40 and then hover over the build in the top right and click "open report"

.travis.yml
```diff
@@ -29,3 +32,5 @@ script:
   - npm run lint:js
   - npm run build --environment production
   - npm test
+  - npx ember build -prod
```
sivakumar-kailasam: Aren't we already building a prod build in couple of steps above?

Maybe in a future PR we could consider splitting these steps as different travis stages so that the build can be run in parallel to the lighthouse analysis.

mansona: I'm probably just a little distracted this week that I didn't notice 🙈 we can delete the earlier one, we just need to make sure that the prod build is the step right before the lighthouse so it runs against the production optimised build

Sure we could do stages but I think it's probably fine to leave it like this for simplicity sake 🤔

mansona: Fixed now 👍

PR #625, opened 2020-03-19
Make ember the together framework

gossi:
Because !

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -3,7 +3,7 @@
 <section aria-labelledby="hero-title" class="bg-shape-boxes-bottom bg-dark">
   <div class="container layout">
     <div class="lg:col-5">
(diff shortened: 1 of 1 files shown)


gossi:
I love the change @MelSumner - sounds really great. I don't know what the label `steering` means in this context, but is this the way forward for this ticket?

mansona:
@gossi because this is such a prominent change to the "tagline" of Ember it likely needs to be discussed at the steering committee meeting 👍 (which happens on a Friday)

gossi:
@mansona I expected that, thanks for the explanation - much appreciated and also to consider it 👍

MelSumner:
This is a great idea! We've discussed it in Steering Committee, and decided that we're going to close this PR while we work through the wording some more- there's some more conversations that we need to have about the shift before we change the website. Stay tuned! 👍

PR #638, opened 2020-04-16
Updating release data to 3.17

lupestro:
The release data on the site for latest and beta has been out of date since 3.16 went LTS, so before 3.18 got as far as a blog post, I figured I'd submit the change for 3.17. Hopefully I got everything right.


locks:
Much obliged 🙏

PR #645, opened 2020-05-10
Fix ember engines guides link

villander:
cc: @locks

### DIFF
--- a/app/templates/learn/index.hbs
+++ b/app/templates/learn/index.hbs
@@ -99,7 +99,7 @@
       </li>
 
       <li>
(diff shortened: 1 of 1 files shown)


mansona:
we really should make sure that the old link works so that we don't drop SEO or break any links 🤔 what method of redirects do you have available in the place that it's currently hosted?

villander:
It already implemented on Engines Guides @mansona 

https://github.com/ember-engines/ember-engines.com/blob/master/tests/dummy/app/routes/guide.js

mansona:
@villander so that's not a "SEO redirect". That won't allow links that are giving any sort of domain authority to apply it to the app any more. I will make sure that I create an issue that covers the problem or maybe we can bring it up in the learning team meeting 👍

villander:
@mansona it was required by @locks to fix SEO. Now I'm not sure what's right.

PR #646, opened 2020-06-01
statement to express community solidarity

MelSumner:
Requesting merge (to revert when appropriate) or temporary branch deployment. 

Copied/Adapted from https://www.django-rest-framework.org/ 

As requested by https://twitter.com/kwuchu/status/1267438396340932610


kategengler:
My only concern is that (from reviewing Percy) this also takes down the Community Guidelines that is linked to from many places and feels important to be available for review.

kategengler:
To clarify, my concern is not linking to the guidelines from the message, but ensuring the URL for the guidelines http://emberjs.com/guidelines still has them available to peruse.

tomdale:
I am in favor of this change.

I was happy to see Ember used at the beginning of the COVID-19 pandemic to quickly roll out a global tracking dashboard, and I hope Ember will similarly be used now in the pursuit of justice.  I believe this PR strikes the right balance of making a bold statement while keeping learning content available for people building applications to hold the powerful accountable.

Second, we serve an international audience and the U.S. does not have a monopoly on injustice. We should use this opportunity to start defining a set of rules for when and how we make these kinds of statements, to avoid sending the signal that only American pain matters.

But that discussion feels too academic to have right at this moment, when my Black brothers and sisters are being murdered by militarized police who are never held accountable for their actions. As small of a gesture this is, I want them to know that we are united in their struggle and that Black Lives Matter.

wycats:
This looks good to me.

sandstrom:
Not against the message, but I'm curious why this issue is singled out. I think you (the US) has a pretty flawed justice system[1][2][3][4] and these flaws has been known for a long time. But it's not just you. 

For example, the Philippine police has been on a fairly well-documented killing spree[5][6][7] during the past few years, in a country where police is killing about ~10x as many per year compared to the US.

Again, I'm not against open-source projects taking a stance. Though personally, I think it would be better to focus on issues closer to the software field, e.g. software patents and government efforts to weaken encryption.

Obviously there is a ton of social and economical injustices that one could get involved in. However, your flawed justice system is actually pretty functional compared to some other places. So if you want to be non-discriminatory in what causes you support, up next should be a 10x longer blackout to support everyone killed by police in the Philippines, and then on to the next[8] one[9].

[1] https://en.wikipedia.org/wiki/Plea_bargain
[2] https://en.wikipedia.org/wiki/Civil_forfeiture_in_the_United_States
[3] https://en.wikipedia.org/wiki/United_States_incarceration_rate
[4] https://en.wikipedia.org/wiki/American_rule_(attorney%27s_fees)
[5] https://www.economist.com/briefing/2020/02/20/rodrigo-dutertes-lawless-war-on-drugs-is-wildly-popular
[6] https://www.theguardian.com/world/2019/jul/05/philippines-senator-dela-rosa-defends-police-over-toddlers-death-in-drugs-raid
[7] https://www.bbc.com/news/world-asia-50236481
[8] https://en.wikipedia.org/wiki/List_of_killings_by_law_enforcement_officers_by_country
[9] https://worldjusticeproject.org/rule-of-law-index/

PR #651, opened 2020-06-08
Explicitly identify projects that have LTS support

GCheung55:
Explicitly identify projects that have LTS support (Ember.js, Ember CLI, Ember Data).


amyrlam:
thanks!

PR #652, opened 2020-06-18
Added Isaac Lee to Team page

ijlee2:
👋


amyrlam:
re-running CI

amyrlam:
With think that the error may be from https://github.com/nodejs/node/issues/32852#issuecomment-613655150, which is in the package-lock.json. It may make sense to do a mass upgrade of packages...if their packages are even up to date. I'm going to ignore the error for now, merge and check the build

PR #655, opened 2020-06-24
Remove releases timeline and replace with links to releases

jenweber:
The timeline is never correct, and even if we automated it to project a date forward from a beta/past minor stable release, there are still no guarantees of releases on specific dates. The Ember project releases _roughly__ every 6 weeks but given the collaborative, cross-team, cross-company nature of the work, human personal reasons, and the seriousness with which we treat stability, it is almost never exactly 6 weeks.

This points people towards the release blog posts as their marker for "roughly every six weeks." Devs could look at our actual blog post release dates to get an idea of the variability.


jenweber:
Screen Shot 2020-06-24 at 12 05 06 AM

jenweber:
For anyone interested, based on release blog post dates, the intervals for this year are:

May 5
    7 weeks 1 day
March 16 
    < 6 weeks
Feb 12 
    7 weeks 5 days
Dec 20
    < 5 weeks
Nov 18
    7 weeks 5 days
Sept 25
    < 6 weeks
Aug 16 
    < 5 weeks
Jul 15 
    7 weeks 6 days
May 21
    5 weeks 6 days
April 10
   6 weeks 1 day
Feb 27
   7 weeks 2 days
Jan 7th

This is pretty ok, given that our goal is to release roughly every 6 weeks! But about 25% of the time, we are over 7 weeks. So, it's not setting good user expectations for us to put actual calendar dates.

app/templates/releases/index.hbs
```diff
@@ -25,7 +25,29 @@
...
 <h2>Upcoming releases</h2>
 
-<ReleaseTimeline class="hide-on-mobile" @project={{emberBetaProject}} />
+The Ember project aims to release approximately every 6 weeks.
+For a very rough estimate of the timeline for the next release,
+you can look at the date of the most recent
+<a href="https://blog.emberjs.com/tags/releases.html">
+  official release blog post
+</a>
+and count forward by six weeks.
```
locks, line 34: This is not accurate. The six week schedule is independent of delays. For a practical example, 3.19 was "released" this week and 3.20 is still scheduled for the week of July 6, two weeks from now.

app/templates/releases/index.hbs
```diff
@@ -25,7 +25,29 @@
...
+The release timelines fluctuate based on team availability
+across all projects, holidays, and for extended testing
+before a release.
```
locks, line 37: This sounds like a repeat of the above. Do you mean to say that the actual releases aren't on the dot because of team availability?

app/templates/releases/index.hbs
jaredgalanis, line 34:
```suggestion
and count forward by 6 weeks.
```
Minor thing, but it feels better to me to be consistent (using numbers vs spelled-out) when referencing the # of weeks. So we'd either change that here and in Our goals or reverse the references using numbers.

jaredgalanis:
@jenweber seems like you've got an import ref to the removed css file that needs to come out and might be breaking the deploy preview.

app/templates/releases/index.hbs
jenweber: More or less, yeah. I can simplify this some!

jenweber:
Thank you @jaredgalanis. Fixed!

@locks could you re-review and let me know what you think? I edited the description to refer to the existence of a schedule.

@ijlee2 I worked in some of your suggestions for the "Create an app using the latest release" section. What do you think? As for the 6 weeks under the goals, that bullet point already existed. Let me know if there was something else you think should be added/changed.

Thanks everyone for the feedback!

jenweber:
Some ideas: 
- put the graphic back but remove the dates so it's not a wall of text

Questions we have
- how to reconcile the schedule with reality? How to communicate the multi-step nature across many projects

PR #656, opened 2020-06-26
Add black-lives-matter route

mansona:
This is the change that was discussed at the Steering Committee 👍


app/components/black-lives-matter-banner.hbs
```diff
@@ -0,0 +1,5 @@
+{{#if this.show}}
+  <div class="survey-reminder-wrapper">
+    {{#link-to 'community.black-lives-matter'}}Black Lives Matter{{/link-to}}
```
ijlee2, line 3: I think we can use the angle bracket syntax in this project now.

```suggestion
    <LinkTo @route="community.black-lives-matter">Black Lives Matter</LinkTo>
```

mansona: 👍

PR #657, opened 2020-06-26
fix asset-size-action

mansona:


mansona:
I'm doing this a different way in #658 👍

PR #658, opened 2020-06-26
Update dependencies

mansona:
This is updating ember-styleguide and ember-leaflet

This is intended to fix the asset-size check because we needed to update `@babel/preset-env`. There is a bit of a catch though, the first time that we actually fix asset-size it will still fail because the current implementation runs it once agains the head of the branch and once against master. The whole check won't succeed until both of those runs succeed 🙈


mansona:
it looks like the asset sizes build has been fixed 🎉 once we merge this the **next** PR will get the asset size reporting again 👍

Issue #672, opened 2020-10-16
Fix JS lint errors


ijlee2:
## Background

When I upgraded lint-related packages to their latest version in https://github.com/ember-learn/ember-website/pull/670, the app didn't meet the latest standards of [eslint](https://eslint.org/docs/rules/) and [eslint-plugin-ember](https://github.com/ember-cli/eslint-plugin-ember). To get the pull request in a mergeable state, I suppressed the lint errors.

## Problem

Let's fix the JS lint errors. 🌟

Warning: I **highly recommend** pushing a commit after you fix 1 _type_ of lint error. Because there are very few application and rendering tests, I also recommend that you run the app locally to check if you may have introduced a bug.

## TODOs

- [ ] Remove (or comment out) 1 of the [six lint checks in `.eslintrc.js`](https://github.com/ember-learn/ember-website/blob/50773b6a94465fa5b723d0b53a3530063c65694a/.eslintrc.js#L34-L39). Run the following command to see where errors occur:

    ```bash
    npm run lint:js
    ```

- [ ] Fix errors, commit the changes, then make a push to GitHub. The CI may help with asserting that the app continues to work.

- [ ] Repeat the two steps above for the 5 remaining lint error types.

- [ ] EmberFest participants: If there are lint errors that you weren't able to fix on time, no worries. Please feel free to submit a pull request with the changes that you were able to make. 🙌🏼

## References

- [ember/no-assignment-of-untracked-properties-used-in-tracking-contexts](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-assignment-of-untracked-properties-used-in-tracking-contexts.md)
- [ember/no-get](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get.md)
- [ember/no-get-with-default](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/no-get-with-default.md)
- [ember/require-computed-property-dependencies](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/require-computed-property-dependencies.md)
- [ember/use-ember-data-rfc-395-imports](https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/use-ember-data-rfc-395-imports.md)
- [no-prototype-builtins](https://eslint.org/docs/rules/no-prototype-builtins)
- Codemod: [ember-data-codemod](https://github.com/ember-codemods/ember-data-codemod)

[...]

PR #678, opened 2020-10-18
Fixed three eslint issues - 

manurampandit:
As part of https://github.com/ember-learn/ember-website/issues/672 , fixing three other linting issues.
what-wg-fetch addition is due to test failing. Reference taken from - https://github.com/ember-cli/ember-fetch/issues/547

Have tested this in browser and running 'ember t' 

@ijlee2 - Kindly review this PR as well.

### DIFF
--- a/app/components/high-charts.js
+++ b/app/components/high-charts.js
@@ -79,14 +79,14 @@ export default Component.extend({
   callback: undefined,
 
   buildOptions: computed('chartOptions', 'content.[]', function() {
-    let theme = this.theme || {};
-    let passedChartOptions = this.chartOptions || {};
+    let theme = this.theme ?? {};
+    let passedChartOptions = this.chartOptions ?? {};
 
     let chartOptions = merge(theme, passedChartOptions);
     let chartContent = this.content;
 
     // if 'no-data-to-display' module has been imported, keep empty series and leave it to highcharts to show no data label.
-    if (!this.content.length && !this.highcharts.Chart.prototype.showNoData) {
+    if (!this.content?.length && !this.highcharts.Chart.prototype.showNoData) {
       chartContent = [{
         id: 'noData',
         data: 0,
@@ -161,7 +161,7 @@ export default Component.extend({
 
   draw() {
     let element = this.element && this.element.querySelector('.chart-container');
-    let modeAttr = this.mode || undefined;
+    let modeAttr = this.mode ?? undefined;
     let mode = CHART_TYPES[modeAttr];
     let completeChartOptions = [this.buildOptions, this.callback];
 
--- a/app/components/high-charts.js
+++ b/app/components/high-charts.js
@@ -1,7 +1,7 @@
 import { assign } from '@ember/polyfills';
 import Component from '@ember/component';
 import { getOwner } from '@ember/application';
-import { set, getProperties, get, computed, getWithDefault } from '@ember/object';
+import { set, computed } from '@ember/object';
 import { run } from '@ember/runloop';
 import merge from 'deepmerge';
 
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -88,6 +88,9 @@
       "lib/content-data-generator"
     ]
   },
+  "resolutions": {
```
mansona, line 91: so this won't work for a number of reasons: 

- this project is not using yarn so resolutions will not have any effect
- we should not be relying on resolutions to fix issues like this
- it seems like this problem will cause issues after upgrading as mentioned in https://github.com/ember-cli/ember-fetch/issues/547#issuecomment-671885942

can you revert this change, and if it is still causing an issue then we can look at upgrading ember-fetch which seems to have fixed the issue 👍

app/routes/application.js
```diff
@@ -25,7 +25,7 @@ export default Route.extend({
     }
 
     const page = this.url;
-    const title = this.getWithDefault('currentRouteName', 'unknown');
+    const title = this.currentRouteName || 'unknown';
```
mansona, line 28: can you please make use of `??` instead of `||`. I know in this case the currentRouteName is supposed to be a string so it will only be falsey if it is empty, but I would prefer to make use of `??` for default values as a standard 👍

app/controllers/community/meetups/index.js
```diff
@@ -10,7 +10,7 @@ export default Controller.extend({
   init() {
     this._super(...arguments);
     
-    if (!this.get('fastboot.isFastBoot')) {
+    if (!this.fastboot.isFastBoot) {
```
mansona, line 13: this has actually subtly changed the semantics of the line. Now if fastboot is null or undefined it will throw an error (where it would not have before).

For us to keep the same semantics you would need to do something like this: 

```suggestion
    if (!this.fastboot?.isFastBoot) {
```

app/components/high-charts.js
```diff
@@ -161,9 +161,9 @@ export default Component.extend({
 
   draw() {
     let element = this.element && this.element.querySelector('.chart-container');
-    let modeAttr = get(this, 'mode') || undefined;
+    let modeAttr = this.mode || undefined;
```
mansona, line 164:
```suggestion
    let modeAttr = this.mode ?? undefined;
```

app/components/high-charts.js
```diff
@@ -79,14 +79,14 @@ export default Component.extend({
...
+    let theme = this.theme || {};
+    let passedChartOptions = this.chartOptions || {};
 
     let chartOptions = merge(theme, passedChartOptions);
-    let chartContent = get(this, 'content');
+    let chartContent = this.content;
 
     // if 'no-data-to-display' module has been imported, keep empty series and leave it to highcharts to show no data label.
-    if (!get(this, 'content.length') && !this.highcharts.Chart.prototype.showNoData) {
+    if (!this.content.length && !this.highcharts.Chart.prototype.showNoData) {
```
mansona, line 89: again this breaks the semantics of what it was doing before: 

This will fix it: 
```suggestion
    if (!this.content?.length && !this.highcharts.Chart.prototype.showNoData) {
```

app/components/high-charts.js
```diff
@@ -79,14 +79,14 @@ export default Component.extend({
   callback: undefined,
 
   buildOptions: computed('chartOptions', 'content.[]', function() {
-    let theme = getWithDefault(this, 'theme', {});
-    let passedChartOptions = getWithDefault(this, 'chartOptions', {});
+    let theme = this.theme || {};
+    let passedChartOptions = this.chartOptions || {};
```
mansona, line 83:
```suggestion
    let passedChartOptions = this.chartOptions ?? {};
```

app/components/high-charts.js
```diff
@@ -79,14 +79,14 @@ export default Component.extend({
   callback: undefined,
 
   buildOptions: computed('chartOptions', 'content.[]', function() {
-    let theme = getWithDefault(this, 'theme', {});
-    let passedChartOptions = getWithDefault(this, 'chartOptions', {});
+    let theme = this.theme || {};
```
mansona, line 82:
```suggestion
    let theme = this.theme ?? {};
```

mansona (changes requested):
Hey thanks for your contribution 🎉 I've added a few comments that will probably need to be looked at before we can get this merged, but overall it's a great job 👍

package.json
manurampandit: Thanks a lot for reviewing this PR. I've updated ember-fetch version as this issue was still occurring.

app/routes/application.js
manurampandit: Thank you for reminding me[ Coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator) use. Almost forgot :)

app/controllers/community/meetups/index.js
manurampandit: Updated this.

manurampandit:
Thank you @mansona for reviewing this and providing qualitative feedback. I've addressed all the comments. Kindly re-review and let me know if I need to change anything else.

mansona (approved):

PR #682, opened 2020-10-29
Remove RFC Tracking section

locks:
Given that the tracking initiative is not really representative of the current state of development and RFC Stages will eventually replace it, I think we should not point our community to it.


MelSumner (approved):
Agree! 👍

mansona (approved):
nice one 👍

PR #685, opened 2020-11-25
Adds new Ember Observer logo with license

MinThaMie:
Discussed with @mansona


ijlee2:
@mansona @MinThaMie Hello. I was wondering what the status of this pull request is. 🙂

mansona (approved):
I think this is good to go but I would appreciate it if someone else gave it a quick look (since I did some work on this PR too)

ijlee2 (changes requested):
From the Percy snapshot and deploy website, I noticed that, on a mobile width, the image no longer stays inside its container. Can we fix the styling issue before merging the PR?

Before:

Screenshot 2021-02-03 at 08 18 24

After:

Screenshot 2021-02-03 at 08 18 41

ijlee2:
I think this might work? When resizing the browser, how the image looks and is positioned seemed similar to what happens on emberjs.com.

```html
<svg
  style="max-width: 460px;"
  viewBox="0 0 460 360"
  xmlns="http://www.w3.org/2000/svg"
  xmlns:xlink="http://www.w3.org/1999/xlink"
>
```

mansona:
First of all: great catch @ijlee2 🎉 I wouldn't have even thought of this being a problem and I'm sorry to say I didn't check the mobile Percy shot 😞 whoops!

second of all: this seems to be a harder problem than I was expecting 🙈 I read through this https://css-tricks.com/scale-svg/ and it seems like the approach that I've picked `Use Inline SVG and the latest Blink/Firefox Browsers` is the thing that is closest to what we want. Let me know what you think and if we can get this merged 👍

ijlee2 (approved):
Looks good! Thanks, @MinThaMie and @mansona, for working on updating the logo.

PR #686, opened 2020-12-02
Update SupportedLTS's

stefanpenner:
According to the support policy, ember@3.12 is no longer in support (as
of October 2020);

I noticed this, as I was implementing "support policy window tooling",
which disagreed with the fixture data harvested from this site. After
investigating, it appears this site's data may be out of date. Or I
misunderstand ember LTS policy.


ijlee2 (approved):
Thanks for updating the LTS page!

Issue #687, opened 2020-12-02
[BUG] LTS Support Verbiage could be improved


stefanpenner:
**Describe the bug**

The site's verbiage around LTS support could be improved slightly. Currently, reading the verbiage would imply the following:

```
1.0.0 <-- LTS (but in-practice, it appear we skip the first of every major as well)
1.1.0
1.2.0
1.3.0
1.4.0 <-- LTS
1.5.0
1.6.0 <-- LTS
2.0.0 <-- LTS (but in-practice, it appear we skip the first of every major as well)
```

This is due to the rules of an LTS being:

* every 4 four minor versions
* plus the final version before a major release

With the addition of the following rule (or a similar rule), the sites verbiage would be inline with how LTS appears to be implemented in practice:

* except the first version of a major release

The sites verbiage for reference:
> An LTS is declared roughly every four minor versions, plus the final version before a major release. Before a version can be called an "LTS" release, it has to spend at least 6 weeks as a stable release, where it is used and tested by thousands of developers.

---

_this seems pedantic, why did this verbiage effect you? This was noticed while implementing tooling to surface unsupported dependencies of a project, when implementing the ember LTS policy I realized my first implementation based on the verbiage has an error, and also surfaced -> https://github.com/ember-learn/ember-website/pull/686_

stefanpenner:
@ijlee2 👍

PR #688, opened 2020-12-02
Clarified LTS schedule

ijlee2:
## Description

This pull request closes https://github.com/ember-learn/ember-website/issues/687. I added an example (Ember 2.x) to illustrate how LTS's may be assigned.

### DIFF
--- a/app/templates/releases/lts.hbs
+++ b/app/templates/releases/lts.hbs
@@ -3,7 +3,7 @@
 <ProjectListing
   @projects={{array @model}}
   @channel="LTS"
-  @channelDescription="Long Term Support releases receive bugfixes and security updates for an extended period."
+  @channelDescription="Long Term Support (LTS) releases receive bugfixes and security updates for an extended period."
 >
 
   <h2>What is an LTS release?</h2>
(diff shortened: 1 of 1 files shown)


ijlee2:
Hi, @stefanpenner. Can you visit https://deploy-preview-688--ember-website.netlify.app/releases/lts and provide feedback for the new description of the LTS schedule?

app/templates/releases/lts.hbs
```diff
@@ -18,10 +18,14 @@
     since LTS versions get even more testing and scrutiny than usual.
   </p>
   <p>
-    An LTS is declared roughly every four minor versions, plus the final version before a
-    major release. Before a version can be called an "LTS" release, it has to spend
-    at least 6 weeks as a stable release, where it is used and tested by thousands of
-    developers.
+    An LTS is declared roughly every 4 minor versions, excluding the x.0 minor version.
```
stefanpenner, line 21: Is their ambiguity here still? I'm not sure if excluding implies that we start counting at x.0 or x.1

x.0
x.1
x.2
x.3 <-- fourth version
x.4 <-- actual LTS

I can be convinced the above isn't ambiguous, but i wonder if something like. Starting after <clause> helps? Ultimately this is still an improvement.

ijlee2: Hmm. I think we may assume that a developer who visits the LTS page is somewhat familiar with semantic versioning. In particular, the developer understands that the major, minor, and patch version numbers begin at 0. (At least, this is my understanding of version numbers.)

So, every 4 minor versions would mean `0, 4, 8, ...`. Afterwards, excluding 0 would result in `4, 8, ...`.

In a subsequent sentence, I provided an example with Ember `2.x` series, in case the pattern may not be still clear to a developer. As a result, I think what's currently written would be okay. What are your thoughts?

ijlee2:
I'll go ahead and merge this PR. Let's open a new PR if we want to revise further. 🙂

Issue #550, opened 2020-02-06
background elements


MelSumner:
See screenshot- see if we can add some background grid dots in the Ember Ecosystem and Batteries included section. 

Not a blocker. 

![image](https://user-images.githubusercontent.com/4587451/73949590-3ba7e100-48c0-11ea-90a6-34157eea6fa0.png)

mansona:
As this is not a blocker I'm going to move it to phase 2 for now 👍

jeffdaley:
@MelSumner Do you have a screenshot of how the "batteries included" section should look?

oo6:
The grid dots cover part of the text and tabs, so can't select or click

<img src="https://user-images.githubusercontent.com/6327995/103846765-c9480e00-50d9-11eb-90d9-14ed3c45789d.png" width="400">

ijlee2:
@oo6 Thank you for reporting the issue!

@jeffdaley Can you create a bug fix PR? I briefly looked at `<GridDots>` component's template, and am not sure yet whether the solution would be to update the `<svg>` definition or to do something else.

jeffdaley:
@oo6 Great catch.

Just submitted a PR addressing this issue.

MinThaMie:
They're there :) So I'm going to close this issue. @jeffdaley thanks for your contribution!

PR #697, opened 2021-01-07
Removes pointer interactivity from grid dots

jeffdaley:
Addresses #550 

`pointer-events: none` seems to be the simplest solution.


ijlee2 (approved):
Thank you for working on the solution quickly! I checked that I can click on all 4 tabs in the preview site (1) on Chrome, Edge, Firefox, and Safari and (2) at a few different resolutions.

PR #706, opened 2021-01-11
Added tests to better document the project (Part 2)

ijlee2:
## Background

Over the next month, I want to make several pull requests so that we can upgrade `ember-source` from `v3.20` to `v3.24`.

## Description

In this pull request, I added unit tests for `application` serializer and default rendering tests for components.

The tests for components, in particular, will help us migrate the components layout from classic pre-Octane (template files in `app/templates`) to classic Octane (colocated in `app/components`).

The default rendering tests also help us check that the code in a backing class can run without error when the arguments are undefined.

### DIFF
--- a/tests/integration/components/high-charts-test.js
+++ b/tests/integration/components/high-charts-test.js
@@ -1,0 +1,16 @@
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+import { setupRenderingTest } from 'ember-qunit';
+import { module, skip } from 'qunit';
+
+module('Integration | Component | high-charts', function (hooks) {
+  setupRenderingTest(hooks);
+
+  skip('it renders', async function (assert) {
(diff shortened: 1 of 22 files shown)


tests/integration/components/high-charts-test.js
```diff
@@ -0,0 +1,16 @@
...
+import { hbs } from 'ember-cli-htmlbars';
+import { setupRenderingTest } from 'ember-qunit';
+import { module, skip } from 'qunit';
+
+module('Integration | Component | high-charts', function (hooks) {
+  setupRenderingTest(hooks);
+
+  skip('it renders', async function (assert) {
+    await render(hbs`
+      <HighCharts />
```
ijlee2, line 11: Rendering the component in a test currently results in the error `calling set on destroyed object`. This is a known issue that, I think, can be better fixed by updating the component to Glimmer.

PR #715, opened 2021-01-12
Installed ember-cli-mirage

ijlee2:
## What Changed

I installed [ember-cli-mirage](https://www.ember-cli-mirage.com/) to help with writing tests.

Previously, I had created fixture data for unit tests for models. In this pull request, I extracted the data into files in `mirage/data` so that we can reuse them for application tests.

## Potential Upsides

We can mock requests (potentially save test runtime?) and control data for test assertions and Percy snapshots. A few Percy snapshots can become [more manageable in terms of image size](https://percy.io/Ember/Ember-Website/builds-next/8342620/changed/474532778?browser=firefox&browser_ids=14&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=375%2C1280) (dimensions).

## Potential Downsides

If we don't use live data, then Percy won't be able to alert us when we make a change to the `data` directory. The directory is updated when we:

- Make an Ember release (`data/project`), approximately every 6 weeks
- Add an Ember.js user (`data/user`)
- Add an Ember core member (`data/team-member`)
- etc.

(listed in descending order of frequency)

We could still use Netlify's deploy preview to check that we correctly made the change.

## How to Review


jenweber (approved):
I think this is a great idea! It is a much better test-writing experience.

PR #717, opened 2021-01-14
Extracted replaceLinks from the application controller

ijlee2:
## Background

## Description

In this pull request, I extracted the function `replaceLinks` from the `application` controller to a utility file called `replace-links`.

This way, we can write unit tests to document what `replaceLinks` is supposed to do, and help a contributor easily rewrite the `application` controller in native class.

## How to Review

I made a few refactors so I recommend reviewing the commits one by one.

### DIFF
--- a/app/utils/replace-links.js
+++ b/app/utils/replace-links.js
@@ -1,0 +1,35 @@
+function replaceUrlPrefix(url) {
+  // ignore external apps
+  if (
+    url.endsWith('/blog') ||
+    url.endsWith('/deprecations') ||
+    url.endsWith('/api')
+  ) {
+    return url;
+  }
+
+  return url
+    .replace(/^https:\/\/emberjs.com\//, '/')
+    .replace(/^\/builds(\/\w+)$/, '/releases$1')
+    .replace(/\/builds$/, '/releases');
+}
+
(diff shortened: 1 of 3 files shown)


app/utils/replace-links.js
```diff
@@ -0,0 +1,44 @@
...
+function replaceInternalLinks(url) {
+  if (isExternalLink(url)) {
+    return url;
+  }
+
+  // Map internal links to the correct route name
+  return url
+    .replace(/^https:\/\/emberjs.com\//, '/')
+    .replace(/^\/builds(\/\w+)$/, '/releases$1')
+    .replace(/\/builds$/, '/releases');
```
ijlee2, line 23: These lines may be no longer needed so consider removing them.

All [header links](https://github.com/ember-learn/ember-styleguide/blob/ab1d1fc32dd023f287c49d3fd700216ba368771a/addon/constants/links.js) and [footer links](https://github.com/ember-learn/ember-styleguide/blob/7a836bda9ce13e3dbddc166b8f56a0425f8eafe6/addon/constants/es-footer.js) that come from `ember-styleguide` do not include the word `builds` in the URL. We also use `.netlifyredirects` to redirect a user if the URL happens to include `builds`.

PR #720, opened 2021-01-16
Added tests to better document the project (Part 7)

ijlee2:
## Background

## Description

In this pull request, I added unit tests for controllers. These tests, in particular, assert how we sort data that are returned from the `model` hook.

Having these tests will help us rewrite controllers in native class. I imagine that, if we were to create components to break down large route templates, we can reuse the test setups that I created in this pull request.

## How to Review

### DIFF
--- a/app/controllers/community/meetups/index.js
+++ b/app/controllers/community/meetups/index.js
@@ -12,6 +12,10 @@ export default Controller.extend({
 
     if (!this.fastboot?.isFastBoot) {
       import('leaflet').then(() => {
+        if (this.isDestroyed || this.isDestroying) {
+          return;
+        }
+
         let prefix = '';
         const config = getOwner(this).resolveRegistration('config:environment');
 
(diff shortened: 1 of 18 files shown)


app/controllers/community/meetups/index.js
```diff
@@ -12,6 +12,10 @@ export default Controller.extend({
 
     if (!this.fastboot?.isFastBoot) {
       import('leaflet').then(() => {
+        if (this.isDestroyed || this.isDestroying) {
+          return;
+        }
```
ijlee2, line 17: I made a temporary fix so that unit tests for `community.meetups.index` controller won't run into `calling set on destroyed object` error.

app/controllers/community/meetups/index.js
```diff
@@ -26,6 +30,6 @@ export default Controller.extend({
   lng: 0,
   zoom: 2,
   meetupsByArea: groupBy('model', 'area'),
-  sortingKeys: Object.freeze(['items.length:desc']),
+  sortingKeys: Object.freeze(['items.length:desc', 'value:asc']),
```
ijlee2, line 33: I added a tiebreaker (sort in ascending order by the area name, which is called `value`).

app/controllers/team.js
```diff
@@ -9,7 +9,7 @@ function inTeam(team) {
 }
 
 export default Controller.extend({
-  sortingKeys: Object.freeze(['added']),
+  sortingKeys: Object.freeze(['last', 'first']),
```
ijlee2, line 12: I changed how we sort people. I thought it's more likely to see people sorted by their name in a website.

PR #729, opened 2021-01-17
Added tests to better document the project (Part 9)

ijlee2:
## Background

## Description

In this pull request, I added application tests (Percy snapshot and accessibility audit) for routes that weren't tested before. In short, we now test every route in application tests.

Previously, there were some issues with `calling set on destroyed object` and Percy snapshots for the survey results pages. I noticed that, right now, they didn't seem to be an issue. If they do continue to cause an issue later, let's address it then.


tests/acceptance/index-test.js
```diff
@@ -35,4 +35,121 @@ module('Acceptance | index', function (hooks) {
 
     assert.hasPageTitle('Ember.js - A framework for ambitious web developers');
   });
+
+  test('A user can use the navigation menu', async function (assert) {
+    /* BEGIN: Set up custom assertions */
```
ijlee2, line 40: Ideally, we would write tests that are readable using `ember-test-selector`.

However, because we use `ember-styleguide`—an external library—adding test selectors may be difficult to do. (I haven't tried yet to see if the addon supports splattributes or adding test selectors in another way.)

I created custom assertions to abstract finding the right HTML elements to test.

ijlee2: This test checks the following code in `app/routes/application.js`:

https://github.com/ember-learn/ember-website/blob/af258ba0a0431aa30cdcb207d76df49ab69e018b/app/routes/application.js#L14

PR #734, opened 2021-01-24
Upgrade development dependencies

ijlee2:
## Description

Before I install `ember-composable-helpers` and `ember-exam`, I wanted to upgrade some of the development dependencies to their latest version.

## Notes

Upgrading `qunit` from `v2.13` to `v2.14` will result in a styling issue when we run `ember test --server`. The issue has been reported in https://github.com/emberjs/ember-qunit/issues/801.

With `qunit v2.13`:

v2 13

With `qunit v2.14`:

v2 14


tests/helpers/page-title.js
```diff
@@ -1,30 +1,15 @@
+import { getPageTitle } from 'ember-page-title/test-support';
```
ijlee2, line 1: The test helper was made public in `ember-page-title v6.2`.

PR #735, opened 2021-01-24
Installed ember-composable-helpers

ijlee2:
## Background

## Description

This pull request involves refactoring code and can be divided into 2 parts.

### Part One (commits 1 - 4)

I installed `ember-composable-helpers` and allowed 3 helpers to simplify controllers. In particular, the `{{sort-by}}` helper can replace the `@sort` method that came from classic Ember ([not recommended in Octane](https://github.com/ember-cli/eslint-plugin-ember/blob/v10.1.2/docs/rules/no-computed-properties-in-native-classes.md)).

### Part Two (commits 5 - 6)

I uninstalled `ember-group-by` because it was used only in 1 place. We can create a native getter instead, and later extract the function into a utility if we want to reuse the function.

## How to Review


app/templates/mascots/index.hbs
```diff
@@ -10,9 +10,8 @@
 
   <Mascots::MascotFilters />
 
-  {{!-- TODO: Use ember-composable-helpers to sort filteredMascots by date:desc --}}
   <Mascots::MascotList
     @display={{this.display}}
-    @mascots={{this.filteredMascots}}
+    @mascots={{sort-by "date:desc" this.filteredMascots}}
```
ijlee2, line 15: This addresses the TODO item that I had remarked in #733.

PR #738, opened 2021-01-24
Renamed team route to teams

ijlee2:
## Description

Due to legacy code reasons (I think), the `app/router.js` file shows the following lines:

```javascript
this.route('team', { path: 'teams' });
this.route('team-redirect', { path: 'team' });
```

When I began working on upgrading the app a few weeks ago, I found these lines to be confusing, especially when paired with the `.netlifyredirects` file.

```
/team       /teams
```

In this pull request, I updated the app's router to follow Ember standards:

```javascript
this.route('team-redirect', { path: 'team' });
this.route('teams');
```

I think the new lines better indicate to a new contributor that `teams` is the definitive choice.


app/utils/replace-links.js
```diff
@@ -19,6 +19,7 @@ function replaceInternalLinks(url) {
   // Map internal links to the correct route name
   return url
     .replace(/^https:\/\/emberjs.com\//, '/')
+    .replace(/^\/team$/, '/teams')
```
ijlee2, line 22: A temporary solution until we update the URLs in `ember-styleguide` and make a patch release.

PR #739, opened 2021-01-24
Octanify components (Part 6)

ijlee2:
## Background

## Description

I replaced `<TeamList>` component with `<Teams::Team>`. The latter component and its subcomponent `<Teams::Team::Member>` improve upon testing and maintenance.


app/templates/teams.hbs
```diff
@@ -3,12 +3,12 @@
 <div class="container">
   <h1 class="text-center">The Team Behind Ember</h1>
 
-  <section class="mb-3" aria-labelledby="team-detail">
```
ijlee2, line 4: I believe the `aria-labelledby` attribute is to be used in conjunction with an `id` on a header. This pattern appears commonly in the project (look for `<section>` elements).

I'm not sure why, here, `aria-labelledby="team-detail"` had been added.

app/templates/teams.hbs
```diff
@@ -1,45 +1,59 @@
...
+
+  <div class="mb-3">
     <p class="text-center">
       Ember is an Open Source project that relies on the tireless support of individual contributors. These are the teams
       that guide the development and instruction of Ember.js.
     </p>
+  </div>
+
+  <section class="mb-3">
```
ijlee2, line 13: Better to separate `mb-3` (margins) from the reusable component.

PR #742, opened 2021-01-24
Replaced luxon with dayjs

ijlee2:
## Description

We can reduce the JS bundle size by ~17 KB (gzipped) if we replace `luxon` with `dayjs`.

We only used `luxon` for `{{format-date-time}}` helper. The helper is well-covered by unit tests and Percy snapshots so we can confidently switch the underlying library.

According to `ember-cli-bundlesize`, the JS bundle size is ~615.8KB (before) and is ~598.3 KB (after).


app/helpers/add-weeks.js
```diff
@@ -1,9 +1,6 @@
 import { helper } from '@ember/component/helper';
+import dayjs from 'dayjs';
 
 export default helper(function addWeeks([initialDate, numWeeks = 0]) {
-  let finalDate = new Date(initialDate);
-
-  finalDate.setDate(finalDate.getDate() + 7 * numWeeks);
-
-  return finalDate;
+  return dayjs(initialDate).add(numWeeks, 'week');
```
ijlee2, line 5: Here, we're using the fact that `dayjs` practices immutability.

PR #743, opened 2021-01-25
Octanify components (Part 7)

ijlee2:
## Background

## Description

I replaced `<HighCharts>` component (classic component) with `<Highcharts>` (Glimmer component). The latter comes with a utility and test helper so that we can better practice separation of concerns and testing.

## How to Review


app/components/high-charts.js
```diff
@@ -1,204 +0,0 @@
...
-        },
-      ];
-    }
-
-    let defaults = { series: chartContent };
-
-    return assign(defaults, chartOptions);
-  }),
-
-  didReceiveAttrs() {
```
ijlee2, line 105: The code in `didReceiveAttr` hook was unnecessary. I think the code was meant to take care of (1) state management for chart rendering and (2) re-rendering the chart should the input data change (the data never changes in our app).

app/components/high-charts.js
```diff
@@ -1,204 +0,0 @@
...
-
-  Highcharts.setOptions(localConfig);
-}
-
-/* Map ember-highcharts modes to Highcharts methods
- * https://api.highcharts.com/class-reference/Highcharts.html
- */
-const CHART_TYPES = {
-  StockChart: 'stockChart',
-  Map: 'mapChart',
```
ijlee2, line 69: I don't think `stockChart` and `mapChart` are valid properties anymore.

app/components/high-charts.js
```diff
@@ -1,204 +0,0 @@
...
-let defaultOptions = {
-  plotOptions: {
-    series: {
-      shadow: false,
-    },
-  },
-
-  global: {
-    timezoneOffset: new Date().getTimezoneOffset(),
-  },
```
ijlee2, line 18: `global.timezoneOffset` has been [deprecated since v6.0.5](https://api.highcharts.com/highcharts/global.timezoneOffset).

app/components/highcharts.hbs
```diff
@@ -0,0 +1,8 @@
+<div
+  class="highcharts-wrapper"
+  data-render-state={{if this.isSettled "settled"}}
+  data-test-highcharts-container
+  {{did-insert this.importHighcharts}}
+  {{did-update this.drawChart this.isHighchartsImported}}
+  {{will-destroy this.destroyChart}}
```
ijlee2, line 7: Modifiers FTW.

app/utils/highcharts.js
```diff
@@ -0,0 +1,31 @@
...
+}
+
+const highchartsOptions = {
+  credits: {
+    enabled: false,
+  },
+
+  plotOptions: {
+    series: {
+      animation: false,
```
ijlee2, line 21: I disabled animation to make writing tests easier.

Allowing animation on _all_ charts also felt unnecessary when the user may see no charts or only 1 chart on the top of their screen when the page loads.

app/utils/highcharts.js
```diff
@@ -0,0 +1,31 @@
...
+import merge from 'deepmerge';
+
+function createChartOptions({
+  chartOptions = {},
+  data = [{ color: '#aaaaaa', data: 0, id: 'placeholder' }],
+  theme = {},
+} = {}) {
+  const themePlusChartOptions = merge(theme, chartOptions);
+
+  return assign({ series: data }, themePlusChartOptions);
```
ijlee2, line 11: I think, previously, `assign` from `@ember/polyfills` had been used instead of `Object.assign` because we needed to support IE 11.

tests/helpers/highcharts.js
```diff
@@ -0,0 +1,15 @@
+import { findAll, waitUntil } from '@ember/test-helpers';
+
+export async function waitUntilAllChartsAreDrawn() {
```
ijlee2, line 3: I wrote this test helper so that, when it's time to take a Percy snapshot, we know that all charts have been drawn. Otherwise, we may see something like the following:

without_all_charts_settled

Since `findAll` always returns an array, we can still use this test helper when there are 0 charts or only 1 chart in the test container.

PR #745, opened 2021-01-26
Adds Browser Support Policy Page

pzuraq:
Adds a browser support policy page, as described in the [new browser support policy RFC](https://github.com/emberjs/rfcs/blob/master/text/0685-new-browser-support-policy.md).

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -56,6 +56,8 @@ Router.map(function () {
     this.route('release');
   });
 
+  this.route('browser-support');
+
   this.route('security');
 
   this.route('sponsors');
--- a/app/router.js
+++ b/app/router.js
@@ -13,6 +13,8 @@ Router.map(function () {
 
   this.route('blog', function () {});
 
+  this.route('browser-support');
+
   this.route('community', function () {
     this.route('black-lives-matter');
     this.route('meetups', function () {
@@ -55,8 +57,6 @@ Router.map(function () {
     this.route('lts');
     this.route('release');
   });
-
-  this.route('browser-support');
 
   this.route('security');
 
--- a/app/templates/browser-support.hbs
+++ b/app/templates/browser-support.hbs
@@ -1,0 +1,151 @@
+{{page-title "Browser Support"}}
+<div class="container layout">
+  <section aria-labelledby="section-browser-support-policy">
+    <h1 id="section-browser-support-policy">Ember.js Browser Support Policy</h1>
+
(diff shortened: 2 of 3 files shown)


app/router.js
```diff
@@ -56,6 +56,8 @@ Router.map(function () {
     this.route('release');
   });
 
+  this.route('browser-support');
```
ijlee2, line 59: Could you place this line between the lines for `blog` and `community`? I'd like to keep the list alphabetized to help with maintenance. 🙂

app/templates/browser-support.hbs
```diff
@@ -0,0 +1,151 @@
...
+  <section aria-labelledby="section-browser-support-policy">
+    <h1 id="section-browser-support-policy">Ember.js Browser Support Policy</h1>
+
+    <p>
+      Ember supports the following major browsers:
+    </p>
+
+    <ul class="list-unstyled layout my-3">
+      <EsCard class="lg:col-2">
+        <h3 class="text-center">Desktop</h3>
```
ijlee2, line 12: I think using `<h3>` right after `<h1>` will fail accessibility audit. Can we use a different tag? (We can always customize styling.)

ijlee2 (changes requested):
Hi, Chris. I checked the deploy preview website and thought the content and layout looked great.

Can you update the PR by:

- Adding 2 application tests (Percy snapshot and accessibility audit) for the `browser-support` route? You can copy-paste the `tests/acceptance/community-test.js` file into `tests/acceptance/browser-support-test.js` file, for example.

I believe the `browser-support` route won't be discoverable unless we provided people a direct link.

If we want the page to show up on the top navigation menu, we'll need to make a pull request in `ember-styleguide` repo and update the addon in this project. (There was a change to the nav menu that I wanted to make, so it may be a good idea to combine both tasks into one PR.)

Screenshot 2021-01-26 at 12 45 00

A potential problem may be that we don't currently use the latest version of `ember-styleguide` (`v5`). We'd need to coordinate with the maintainers of `ember-styleguide` to see if they can make a minor release in `v4`.

pzuraq:
Awesome, yep, sounds good 😄  I wasn't sure what the procedure was for adding top level links so I figured we could figure it out. Will update with the changes, and should I make a PR to `ember-styleguide`?

ijlee2 (approved):
Thank you again for helping out with describing the new Browser Support policy!

PR #756, opened 2021-02-08
revert change to yarn

mansona:
This app was accidentally converted to yarn in https://github.com/ember-learn/ember-website/commit/5a3a97d98cbdd6f9e8d4766e729ff510a01ad48e and this just reverts that

I have talked to @ijlee2 about it last week and we agreed that we should get this done 👍


.github/workflows/ci.yml
```diff
@@ -25,26 +25,29 @@ jobs:
...
+        id: npm-cache-dir-path
+        run: echo "::set-output name=dir::$(npm config get cache)"
+
+      - name: Cache npm cache and node_modules
         id: cache-dependencies
         uses: actions/cache@v2
         with:
           path: |
-            ${{ steps.yarn-cache-dir-path.outputs.dir }}
+            ${{ steps.npm-cache-dir-path.outputs.dir }}
```
ijlee2, line 40: [Before the change](https://github.com/ember-learn/ember-website/pull/698/files#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fL33), we just had `~/.npm` (a constant). Can you check if we can do without and remove lines 31-35?

.github/workflows/ci.yml
```diff
@@ -60,28 +63,31 @@ jobs:
...
+        run: echo "::set-output name=dir::$(npm config get cache)"
 
-      - name: Cache Yarn cache and node_modules
+      - name: Cache npm cache and node_modules
         id: cache-dependencies
         uses: actions/cache@v2
         with:
           path: |
-            ${{ steps.yarn-cache-dir-path.outputs.dir }}
+            ${{ steps.npm-cache-dir-path.outputs.dir }}
```
ijlee2, line 78: Similarly to above, can you check if we can write just `~/.npm` and remove lines 66-72?

ijlee2 (approved):
Before merging the PR, can you check if we can simplify the workflow by writing `~/.npm`?

.github/workflows/ci.yml
```diff
@@ -60,28 +63,31 @@ jobs:
         with:
           node-version: ${{ env.NODE_VERSION }}
 
-      - name: Get Yarn cache path
-        id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Update npm
+        run: npm i -g npm@latest
```
rwjblue, line 67: FWIW, I wouldn't do this...

mansona: Ah yes we've discussed this before, I'll make it `npm@7` 👍

.github/workflows/ci.yml
mansona: so I just double checked the documentation that I took this from and it recommends that we use the structure that we have now https://github.com/actions/cache/blob/main/examples.md#node---npm 

I think it's probably best to go with the recommended config 👍

PR #759, opened 2021-02-10
Update mascot-filters.hbs

MelSumner:
If merged, this would add an aria-label to the mascot filter navigation (which is not the only nav on the page, and should therefore have a unique label).


ijlee2 (approved):
Thank you for improving the accessibility of the Ember.js website!

PR #763, opened 2021-02-18
chore: update the deprecations app showcase image

Windvis:
This updates the image with the new design of the deprecations app.

I took a screenshot at 1536x960 and rescaled it to the current 512x320 resolution.


ijlee2 (approved):
Nice, thank you for updating the screenshot!

PR #764, opened 2021-02-18
chore: update the api docs showcase image

Windvis:
The updated image has the following benefits:
- screenshot of a newer design
- image resolution is consistent with the other showcase images
- image size is reduced by using a smaller image

I took a screenshot at 1536x960 and rescaled it to the 512x320 resolution used by the other showcase images.


ijlee2 (approved):
💯

PR #766, opened 2021-02-19
Hide the "Upcoming releases" title on mobile devices

Windvis:
This hides the title on mobile devices and makes it consistent with the Tomster statues.


ijlee2 (approved):
Good catch, I didn't realize `<Releases:ReleaseTimeline>` used `.hide-on-mobile` to be hidden on mobile. Thank you for addressing the issue!

PR #768, opened 2021-02-20
Created PieChart class

ijlee2:
## Description

I would like to start refactoring the code in `ember-community-survey` controller files. As a proof of concept, I created `PieChart` class, which can replace 2 instances of hard-coded chart options.

I created a class that accepts a minimum set of arguments. By creating a class, the controllers don't need to be aware that Highcharts is used to create charts, or that a future major version of Highcharts may require a different configuration. By creating a class, we can also write unit tests for what the configuration object should look like.

My long-term goal is to write a class for every chart type that we have used for the survey results pages. The chart types are:

- areaspline (4 instances)
- bar (6)
- column (36)
- columnrange (1)
- pie (2)
- spline (8)

### DIFF
--- a/app/controllers/ember-community-survey-2020.js
+++ b/app/controllers/ember-community-survey-2020.js
@@ -1,4 +1,5 @@
 import Controller from '@ember/controller';
+import { PieChart } from 'ember-website/utils/highcharts';
 
 // 2020 Color Palette
 // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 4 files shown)


app/controllers/ember-community-survey-2020.js
```diff
@@ -337,32 +338,16 @@ const SS_Q305 = {
...
-      type: 'pie',
-      colors: [emberColors.blue, emberColors.ember],
-    },
-    title: {
-      text: 'Do you internationalize your applications?',
-    },
-    subtitle: { text: '' },
-    tooltip: { pointFormat: '{point.y:.2f}%' },
-    xAxis: { type: 'category' },
-    yAxis: { title: { text: 'Percent of Responses' } },
```
ijlee2, line 352: I think these may be ignored in pie charts. I wasn't able to find the text `Percent of Responses` in the HTML, for example.

app/controllers/ember-community-survey-2020.js
```diff
@@ -337,32 +338,16 @@ const SS_Q305 = {
...
+const SS_Q306 = new PieChart({
+  chart: {
+    title: 'Do you internationalize your applications?',
   },
-  data: [
-    {
-      showInLegend: false,
-      name: '2020',
```
ijlee2, line 357: I think `showInLegend` and `name` may be ignored in pie charts.

MelSumner:
Do we need to do this? I thought it was already really straightforward to use the highcharts code. What benefit does this change bring?

MelSumner (approved):

ijlee2:
Hi, @MelSumner. Thanks for reviewing the PR and asking the question if we need the code changes (it's a good question).

I think our current controller code has 2 characteristics:

- Easy to create (copy-paste existing chart configuration objects)
- Not easy to maintain

Here, I think maintainability refers to the following technical risk: If the object's shape is wrong (e.g. because we didn't create the object correctly, or because a new release of Highcharts asks that we update the object's shape), then we'll need to make many changes of the same type.

Returning to the first characteristic, copy-pasting large objects can feel simple. However, I think it may be good to define a minimal object that will help us (reviewers and new contributors to survey pages) understand why we are creating an object in a certain way.

I hope these thoughts made sense. If you have further questions, please feel free to ask.

PS. One thing that I wasn't sure: How had the configuration objects been created? Was there a script that read the survey results and generated a controller file?

Issue #773, opened 2021-02-24
General Branding Updates


wifelette:
Putting this here so we can all track the todos. 

**Tasks @pichfl is taking care of**:

- [x] Send @wifelette vectors of all the updated Tertiary Logo designs
- [x] Update the not-yet-released `TypedEmber` logo to match the new Tertiary design
- [x] Update the Brand page with all these new assets (#774)
- [x] Fix the weirdness with the _Naming_ section headline on the Brand page

**Tasks @jeffdaley is taking care of**:

- [x] Update the Brand Guide PDF to include the new Tertiary logos, as well as the new Glimmer and Embroider logo sections
- [x] Update the Brand page with a new section for the Glimmer and Embroider logos
- [x] Update the logo on the Glimmer repo/org once the art is ready

**Tasks @mansona can hopefully take care of**:

- [ ] The updated Tertiary logos will need to be shared with any stakeholders for the various projects so their websites, assets, etc. can be updated accordingly

**Tasks @ef4 will take care of**:

- [x] Update the logo on the [Embroider repo/org](https://github.com/embroider-build) once the art is ready

jeffdaley:
Symbol for Embroider org:
![Embroider Symbol](https://user-images.githubusercontent.com/754957/109187583-08045700-7760-11eb-8fcd-2491f638d925.png)

ef4:
Embroider org is now updated:
<img width="554" alt="Screen Shot 2021-03-01 at 2 26 14 PM" src="https://user-images.githubusercontent.com/319282/109548010-18873b00-7a9a-11eb-928d-afc8c05c6d11.png">

wifelette:
@mansona @pichfl This issue is complete in terms of my, Ed and Jeff's tasks :) You can potentially separate out the two remaining items into other issues, or just take care of them, but reminder!

pichfl:
@mansona @wifelette I forgot to ask about this, but what is broken about the headline? It does look normal to me.

jeffdaley:
Whoops, I took care of that one, just forgot to check it off.

mansona:
@wifelette I auto-closed this as I just completed the last item in it, I hope that's ok 😄

PR #774, opened 2021-02-24
Update Ternary logos

mansona:
This adds the new versions of the tertiary logos to the brand page as described in https://github.com/ember-learn/ember-website/issues/773 

A few things to note: 

- I have exported SVGs for the ternary logos and this page now uses those in the examples
- I have also exported PNGs in 1x 2x and 3x sizes for each of the ternary logos and they are included in this PR but they are not accessible from any link on the web 🤔  I notice that ternary logos are not included in the linked `logos-ember.zip` which might seem like a good place to make these PNG versions available but I don't want to add them if they weren't there before
- I have had to make a change to the page so that the "boxes" that demo the ternary logos don't come with padding. This is because these new logos include the correct safe space **inside** the logo so no external padding is required. This also helps us to make sure that we have a consistent padding wherever they are used
- I have deleted the Ember Jobs logo from this page (and the files) as it is no longer a thing


wifelette (approved):
Looks good to me! Added @jeffdaley too since he'll be working on next steps.

> I have also exported PNGs in 1x 2x and 3x sizes for each of the ternary logos and they are included in this PR but they are not accessible from any link on the web 🤔 

If we put _all_ our logos in one file it's a) potentially large but b) likely more than most people need. So maybe we make two links there, or even three, where each is a zipped file of that _category_ of logo?

Good catch on removing the Jobs logo!

jeffdaley:
Looks good.

The SVGs aren't loading correctly for me in Illustrator or Sketch, only in Figma. Probably something to look into.

We also ship a `-small` version of every logo that uses a relatively larger ® symbol. If you look at the logo zip files, you'll see a references for that.

I'm OK merging this and adding the logos to the ZIP in a separate PR.

jeffdaley (approved):

PR #777, opened 2021-02-25
Update branding

jeffdaley:
Loop-based logos
- Places logo content in the `logos` controller so it can be looped through
- Adds intrinsic padding to logo files themselves
- Establishes a consistent computed-file-name pattern

TODO
- [ ] Package all files into ZIP
- [ ] Update guidelines
- [ ] Add typed ember logo


jeffdaley:
Wondering out loud:
Would it be easier to `svg-jar` these logos to reduce HTTP requests?

mansona:
Hey @jeffdaley, just to answer your "wondering out loud" question 😂 

If we svg-jar'd the logos then they would become part of the Javascript bundle (because that's where all templates go at the moment) which could affect any page-speed metrics. It might be a good tradeoff if we were using logos on every page and it didn't matter that we were "making the bundle larger" because there didn't exist any pages that didn't need these logos.

As they are only on one page it's totally fine to have HTTP requests to get them 👍 If we wanted to be really smart we could implement some headers to start pre-loading the SVGs before the HTML has been parsed but it's probably overkill 😂 things like HTTP2 make multiple requests like this much less costly  and we're using that already (thanks Netlify 🎉 )

jeffdaley:
Got it. Thanks for the insight!

PR #780, opened 2021-03-01
Update ember-community-survey-2021.hbs

MelSumner:
Adds the link for the 2021 community survey (to go live March 1, 2021)

Changes to this year (compared to last year): 
- updated for 2021 (prose, Ember versions)
- some instances of checkboxes were changed to radio buttons
- an a11y-related question was removed
- some instances where "how many" selections were overlapping were fixed
- some question answer options were changed, based on last year's results
- more single choice selections were set to be randomized (each user will see the options in a random order)
- additional "other" selections are allowed

Note: there will still need to be some decision made about the website banner. Do we want to change the link that is there, since that's where the survey link has gone in past years? 

Also it would be good (cc @mansona) if emberjs.com/survey could redirect to the survey link for the current year!

### DIFF
--- a/app/templates/ember-community-survey-2021.hbs
+++ b/app/templates/ember-community-survey-2021.hbs
@@ -11,12 +11,12 @@
   <div class="mt-3">
     <a
       class="es-button"
-      href="#"
+      href="https://tilde.wufoo.com/forms/2021-emberjs-community-survey"
       rel="noopener noreferrer"
       target="_blank"
       disabled
     >
-      Coming Soon!
+      Take The Survey Now
     </a>
   </div>
 </EmberCommunitySurvey::Introduction>
--- a/app/templates/ember-community-survey-2021.hbs
+++ b/app/templates/ember-community-survey-2021.hbs
@@ -50,11 +50,11 @@
         >
 
         <h3>
-          Open until March 12th
+          Open until March 14th
         </h3>
 
         <p>
-          Submissions close on March 12th. To ensure the survey captures a fair measure of our community, please share this page on your social networks and with your local communities!
+          Submissions close on March 14th. To ensure the survey captures a fair measure of our community, please share this page on your social networks and with your local communities!
         </p>
       </div>
 
--- a/app/templates/ember-community-survey-2021.hbs
+++ b/app/templates/ember-community-survey-2021.hbs
@@ -14,7 +14,6 @@
       href="https://tilde.wufoo.com/forms/2021-emberjs-community-survey"
       rel="noopener noreferrer"
       target="_blank"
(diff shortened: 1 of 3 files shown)


jenweber (approved):
Looks great! Thank you Melanie!

app/templates/ember-community-survey-2021.hbs
```diff
@@ -11,12 +11,12 @@
   <div class="mt-3">
     <a
       class="es-button"
-      href="#"
+      href="https://tilde.wufoo.com/forms/2021-emberjs-community-survey"
```
ijlee2, line 14: To confirm, line 17 shows that we're using `disabled` attribute. Should this line be removed now that we have a link?

ijlee2 (approved):
@MelSumner We can update `app/templates/application.hbs` to:

```handlebars
<BlackLivesMatterBanner />

<CtaEmberCommunitySurvey
  @surveyRoute="ember-community-survey-2021"
  @surveyTitle="2021 Ember Community Survey"
/>

<EsHeader
  ...
```

I think it's okay if, while the survey is available to take, we temporarily comment out `<BlackLivesMatterBanner />` so that we don't see two orange banners. We can show both, but maybe we'll want to add a visual separation.

Screenshot 2021-03-01 at 07 54 50

ijlee2:
I pushed a commit to add a Percy snapshot and make axe checks in `ember-community-survey-2021` page. Hope that's okay!

app/templates/ember-community-survey-2021.hbs
MelSumner: yes, thanks for catching that

MelSumner:
> @MelSumner We can update `app/templates/application.hbs` to:
> 
> ```
> <BlackLivesMatterBanner />
> 
> <CtaEmberCommunitySurvey
>   @surveyRoute="ember-community-survey-2021"
>   @surveyTitle="2021 Ember Community Survey"
> />
> 
> <EsHeader
>   ...
> ```
> 
> I think it's okay if, while the survey is available to take, we temporarily comment out `<BlackLivesMatterBanner />` so that we don't see two orange banners. We can show both, but maybe we'll want to add a visual separation.
> 
> Screenshot 2021-03-01 at 07 54 50

It's too much to have two banners. I'll chat in Discord, so we can sort it out.

app/templates/ember-community-survey-2021.hbs
```diff
@@ -11,12 +11,12 @@
...
       rel="noopener noreferrer"
       target="_blank"
       disabled
```
MelSumner, line 17:
```suggestion
```

PR #784, opened 2021-03-02
Created AreaSplineChart class

ijlee2:
## Description

A continuation of #778. In this pull request, I created `AreaSplineChart` class, which can replace 3 instances of hard-coded chart options. (A 4th instance in 2020 survey results turned out to be a copy from 2019. I will remove the duplicate code in a separate issue.)

I created a class that accepts a minimum set of arguments. By creating a class, the controllers don't need to be aware that Highcharts is used to create charts, or that a future major version of Highcharts may require a different configuration. By creating a class, we can also write unit tests for what the configuration object needs to look like.

## How to create a chart class

The first 4 commits show step-by-step instructions for creating the `AreaSplineChart` class. I think these steps can help a new contributor with creating additional chart classes and maintaining existing ones.


app/controllers/ember-community-survey-2020.js
```diff
@@ -28,7 +32,7 @@ const orangePalette = {
 // Default Color Assignments for Yearly Graph Data Series
 var color2017 = darkGrayColor;
 var color2018 = emberColors.gray;
-var color2019 = orangePalette.blue;
```
ijlee2, line 31: `orangePalette.blue` does not exist.

PR #785, opened 2021-03-02
Removed unused code (Part 2)

ijlee2:
## Description

While working on #784, I realized that several configuration objects were copy-pastes from 2019, and without additional data for 2020. Since they don't have data for 2020, let's remove them so that we can reduce bundle size and the amount of code that we need to maintain.

I also marked configuration objects in 2019 and 2020 that weren't analyzed and rendered in the template, but did have data for that year.

### DIFF
--- a/app/controllers/ember-community-survey-2020.js
+++ b/app/controllers/ember-community-survey-2020.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-unused-vars */
 import Controller from '@ember/controller';
 import {
   AreaSplineChart,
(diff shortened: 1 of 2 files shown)


app/controllers/ember-community-survey-2020.js
```diff
@@ -269,35 +243,6 @@ const SS_Q302 = {
...
-    tooltip: { pointFormat: '{point.y:.2f}%' },
-    xAxis: { type: 'category' },
-    yAxis: { title: { text: 'Percent' } },
-  },
-  data: [
-    {
-      name: '2019',
-      color: color2019,
-      data: [
-        { name: 'More', y: (100 / 1006) * 530 },
```
ijlee2, line 292: For some reason, the denominator `1006` in the year 2020 didn't match the denominator `1232` in the year 2019, even though the numerators (the actual counts) did match.

Not sure why we "lost" some number of survey responses. This happened in several other places in the 2020 file.

PR #786, opened 2021-03-05
Remove blog route

bertdeblock:
It seems this route is not needed anymore because of the Netlify redirect? Feel free to close if I'm mistaken!

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -11,8 +11,6 @@ Router.map(function () {
     this.route('legal');
   });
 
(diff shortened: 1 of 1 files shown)


ijlee2 (approved):
Yep, in the deploy preview, when I typed `/blog` at the end of the URL, I was redirected to `blog.emberjs.com` still. I think we can try it out live and revert the PR if it doesn't work.

Thanks for the pull request!

PR #787, opened 2021-03-05
Add not-found route

bertdeblock:
Far from perfect, but at the moment non-existent URLs return a blank screen.

Let me know if I should update anything related to UI, UX, A11Y, copy and so on.

Screenshot 2021-03-05 at 19 39 39 (2)

### DIFF
--- a/app/router.js
+++ b/app/router.js
@@ -52,6 +52,8 @@ Router.map(function () {
     this.route('payment-sent');
   });
 
+  this.route('not-found', { path: '/*path' });
+
   this.route('releases', function () {
     this.route('beta');
     this.route('canary');
--- a/app/routes/not-found.js
+++ b/app/routes/not-found.js
@@ -1,0 +1,14 @@
+import Route from '@ember/routing/route';
+import { inject as service } from '@ember/service';
+
+export default class NotFoundRoute extends Route {
+  @service fastboot;
+
+  beforeModel() {
+    if (!this.fastboot.isFastBoot) {
+      return;
+    }
+
+    this.fastboot.response.statusCode = 404;
+  }
+}
--- a/app/templates/not-found.hbs
+++ b/app/templates/not-found.hbs
@@ -1,0 +1,6 @@
+{{page-title "Page Not Found"}}
+
+<div class="container">
+  <h1>Page Not Found</h1>
+  <p>We couldn't find what you were looking for.</p>
+</div>
--- a/tests/acceptance/not-found-test.js
+++ b/tests/acceptance/not-found-test.js
@@ -1,0 +1,32 @@
+import { visit } from '@ember/test-helpers';
+import percySnapshot from '@percy/ember';
+import { a11yAudit } from 'ember-a11y-testing/test-support';
+import { setupMirage } from 'ember-cli-mirage/test-support';
+import { setupApplicationTest } from 'ember-qunit';
+import loadDefaultScenario from 'ember-website/mirage/scenarios/default';
+import { setupPageTitleTest } from 'ember-website/tests/helpers/page-title';
+import { module, test } from 'qunit';
+
+module('Acceptance | not found', function (hooks) {
+  setupApplicationTest(hooks);
+  setupMirage(hooks);
+  setupPageTitleTest(hooks);
+
+  hooks.beforeEach(function () {
+    loadDefaultScenario(this.server);
+  });
+
+  test('Percy snapshot', async function (assert) {
+    await visit('/foo-bar-baz');
+    await percySnapshot(assert);
+
+    assert.ok(true);
+  });
+
+  test('Accessibility audit', async function (assert) {
+    await visit('/foo-bar-baz');
+    await a11yAudit();
+
+    assert.hasPageTitle('Page Not Found - Ember.js');
+  });
+});
--- a/tests/unit/routes/not-found-test.js
+++ b/tests/unit/routes/not-found-test.js
@@ -1,0 +1,12 @@
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+
+module('Unit | Route | not-found', function (hooks) {
+  setupTest(hooks);
+
+  test('it exists', function (assert) {
+    const route = this.owner.lookup('route:not-found');
+
+    assert.ok(route);
+  });
+});


app/templates/not-found.hbs
```diff
@@ -0,0 +1,6 @@
+{{page-title "Page Not Found"}}
+
+<div class="container text-center">
```
ijlee2, line 3: Can we make the page left-aligned?

```suggestion
<div class="container">
```

app/routes/not-found.js
```diff
@@ -0,0 +1,3 @@
+import Route from '@ember/routing/route';
+
+export default class NotFoundRoute extends Route {}
```
ijlee2, line 3: I didn't know this till today. I had a hunch that we would need to return an actual status code of 404 to the user. Since this project uses FastBoot, I think this is how we do so:

```suggestion
import Route from '@ember/routing/route';
import { inject as service } from '@ember/service';

export default class NotFoundRoute extends Route {
  @service fastboot;

  beforeModel() {
    if (!this.fastboot.isFastBoot) {
      return;
    }

    this.fastboot.response.statusCode = 404;
  }
}
```

ijlee2: If you open Chrome's Network tab, you will see the following.

Before:

Screenshot 2021-03-06 at 08 15 52

After:

Screenshot 2021-03-06 at 08 20 26

(If you click on Chrome's Console tab, you will now see the error message `GET http://localhost:4200/foo-bar-baz 404 (Not Found)` as well.)

I learned the information from:

- https://ember-fastboot.com/docs/user-guide#status-code
- https://discuss.emberjs.com/t/404-when-route-is-known-but-model-is-missing

app/router.js
```diff
@@ -52,6 +52,8 @@ Router.map(function () {
     this.route('payment-sent');
   });
 
+  this.route('not-found', { path: '/*path' });
```
ijlee2, line 55: I'm not sure if it matters. I thought one needed to list the 404 route as the very last route in `app/router.js` (since it's a catch-all).

I ran your branch locally and tried to find a failing case if we continue to list the 404 route somewhere in-between. However, I couldn't find a case.

tests/acceptance/not-found-test.js
```diff
@@ -0,0 +1,32 @@
...
+    assert.ok(true);
+  });
+
+  test('Accessibility audit', async function (assert) {
+    await visit('/foo-bar-baz');
+    await a11yAudit();
+
+    assert.hasPageTitle('Page Not Found - Ember.js');
+  });
+});
```
ijlee2, line 32: I wanted to add a test that checks that the status code of 404 is returned. I didn't figure out to, however, and didn't find information online or on Discord. Would you like to give it a try?

If it takes too much time to figure out, I think we can assume that FastBoot will correctly set the status code.

tests/unit/routes/not-found-test.js
```diff
@@ -0,0 +1,12 @@
+import { setupTest } from 'ember-qunit';
+import { module, test } from 'qunit';
+
+module('Unit | Route | not-found', function (hooks) {
+  setupTest(hooks);
+
+  test('it exists', function (assert) {
+    let route = this.owner.lookup('route:not-found');
```
ijlee2, line 8: Just for consistency with other route test files.

```suggestion
    const route = this.owner.lookup('route:not-found');
```

ijlee2 (commented):
Thanks for addressing the lack of a not-found page in the Ember.js website. I think the solution that you took is good, and left a few suggestions for improvement. Will you have a look?

app/router.js
bertdeblock: I would be surprised if it does, I couldn't find anything in the docs as well. I added the route here to keep everything sorted alphabetically. I can still move it towards the bottom if you want? Just to be sure.

tests/acceptance/not-found-test.js
bertdeblock: Added `ember-cli-fastboot-testing` for this. Hope that's okay. It had to be different test though.

app/routes/not-found.js
bertdeblock: TIL 😄

bertdeblock:
Looking into the failing checks. Could they be related to the lock file changes? I also had some lock files changes when installing all modules for the first time.

bertdeblock:
This seems to be the issue we're running into:
https://github.com/ef4/ember-auto-import/issues/340

This PR seems to fix it:
https://github.com/embermap/ember-cli-fastboot-testing/pull/434

app/router.js
ijlee2: I prefer listing routes alphabetically to help with maintenance. If we later find that a URL is incorrectly considered as `not-found`, we can open an issue then.

tests/acceptance/not-found-test.js
ijlee2: Yep, sounds good.

tests/fastboot/not-found-test.js
```diff
@@ -0,0 +1,14 @@
...
+
+module('FastBoot | not-found', function (hooks) {
+  setup(hooks);
+
+  test('it returns the correct status code', async function (assert) {
+    const visitHome = await visit('/');
+    assert.equal(visitHome.statusCode, 200);
+
+    const visitFoo = await visit('/foo');
+    assert.equal(visitFoo.statusCode, 404);
```
ijlee2, line 12: Let me add descriptions to help with maintenance and destructure the returned object to better match the [documentation example](https://embermap.github.io/ember-cli-fastboot-testing/docs/status-code).

```suggestion
    let { statusCode } = await visit('/');

    assert.strictEqual(
      statusCode,
      200,
      'The user sees 200 status code when they visit a URL that exists.'
    );

    ({ statusCode } = await visit('/foo-bar-baz'));

    assert.strictEqual(
      statusCode,
      404,
      'The user sees 404 status code when they visit a URL that does not exist.'
    );
```

ijlee2:
@bertdeblock Ah, I think I see what you meant by failing checks (Netlify deploy logs are showing errors). Yeah, if installing `ember-cli-fastboot-testing` caused the issue, let's skip on adding the status code test and uninstall the addon for now.

Once the https://github.com/embermap/ember-cli-fastboot-testing/pull/434 is merged, we can try installing the addon and adding the test back.

bertdeblock:
@ijlee2 Removed the fastboot test for now. Will create e separate PR for the test.

ijlee2 (approved):
Looks good, thank you for working on this!

PR #789, opened 2021-03-06
Improve LTS table alignment

bertdeblock:
I felt like the readability/alignment of the LTS table could slightly be improved. Hope this is okay.

Before:
Before

After:
After

### DIFF
--- a/app/styles/components/lts-table.css
+++ b/app/styles/components/lts-table.css
@@ -1,0 +1,17 @@
+.lts-table {
+  border-spacing: 0;
+}
+
+.lts-table th {
+  text-align: left;
+}
+
+.lts-table td,
+.lts-table th {
+  padding: 0 var(--spacing-2) var(--spacing-1) 0;
+}
+
+.lts-table td:last-child,
+.lts-table th:last-child {
+  padding-right: 0;
+}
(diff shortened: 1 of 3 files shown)


amyrlam:
Seems like a good improvement to me! 

I remember @locks talking about this section so wanted to tag before merging in, what do you think?

bertdeblock:
Making the names of the months shorter and aligning the dates to the right also seems like a good option to me:
Screenshot 2021-03-07 at 08 00 24

app/styles/components/lts-table.css
```diff
@@ -0,0 +1,17 @@
...
+  border-spacing: 0;
+}
+
+.lts-table th {
+  text-align: left;
+}
+
+.lts-table td,
+.lts-table th {
+  padding: 0 1em 0.5em 0;
```
ijlee2, line 11: Can we use `rem` instead? I think `rem` is always relative to the base font size, whereas `em` is relative to the font size in a parent container (figuring out which container can be difficult, I think).

```suggestion
  padding: 0 1rem 0.5rem 0;
```

ijlee2 (approved):

app/styles/components/lts-table.css
bertdeblock: Used the spacing vars, didn't notice these until now.

amyrlam (approved):
thanks both - lgtm!

PR #802, opened 2021-03-12
Make sure we don't downgrade the lockfileVersion

mansona:
It looks like the package-lock.json file was accidentally downgraded as part of https://github.com/ember-learn/ember-website/pull/769 

This MR adds https://github.com/mansona/npm-lockfile-version which is a super simple check to make sure that this doesn't happen again 👍  (and it also fixes the lockfile so that it now passes the check)


jaredgalanis (approved):
LGTM 👍

PR #803, opened 2021-03-12
[WIP] Update host so that images will show up on Twitter

amyrlam:
Relates to: same change done on ember-blog https://github.com/ember-learn/ember-blog/pull/873

I'll remove WIP if this works on Netlify, had some issues with localhost

PR #804, opened 2021-03-12
Hardcode full path to social image to match ember-blog

amyrlam:
Relates to: https://github.com/ember-learn/ember-website/pull/803 and https://github.com/ember-learn/ember-blog/pull/873

//

I'm not sure if this is the best approach (was originally trying to add the host so that the relative URL wouldn't be `/images/...`), but will sync us up with the blog and look better on Twitter for now. Let me know what you think.

May make sense to have a separate repo for shared assets?

WAS:
image

IS:
image


amyrlam:
Percy error seems unrelated and ok to ignore.

jaredgalanis (approved):
LGTM 👍

jaredgalanis:
> Percy error seems unrelated and ok to ignore.

Hadn't seen your comment, so I approved the diff, but I think it was fine. It's going to be one of those transient diffs that reappears I think.

amyrlam:
thanks! will check tweets after the prod build

PR #805, opened 2021-03-17
add ability to link to a particular mascot

mansona:
I found myself wanting to link to the Empress mascot recently so I added the ability to link to a particular mascot using an ID. 

Once this PR is merged you will be able to go to https://emberjs.com/mascots/#empress and it should scroll to the right part of the page and highlight the "target" mascot like this: 

Screenshot 2021-03-17 at 15 16 19

Let me know if you have any questions :) 

Edit: here is a link with the preview app to allow you to test if it's working https://deploy-preview-805--ember-website.netlify.app/mascots/#empress

Second Edit: this functionality was added in https://github.com/ember-learn/ember-website/pull/1009 and this PR now just adds the highlight to the linked Mascot

### DIFF
--- a/app/components/mascots/mascot-list.hbs
+++ b/app/components/mascots/mascot-list.hbs
@@ -1,6 +1,6 @@
 <ul class="unstyled grid {{if (eq @display "small") "sm:grid-2 lg:grid-3" "sm:grid-1 lg:grid-2"}}">
   {{#each @mascots as |mascot|}}
-    <li id={{mascot.id}} data-test-mascot>
+    <li data-test-mascot>
       <Mascots::MascotList::Item
         @mascot={{mascot}}
       />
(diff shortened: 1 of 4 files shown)


ijlee2 (changes requested):
Hi, @mansona. At the moment, I'm not sure what benefits jumping to a particular image in the Mascots page brings. Could you explain how you need to use this feature?

Fragment identifiers present a few problems:

1. We can't change image file names to improve maintenance (my understanding is that the file name results in the ID), without additional code for redirects.

2. When visiting the URL with a fragment identifier, the page can still scroll to the top due to logic in `application` route.

3. The code change currently ignores accessibility. The focus is on top of the page, not where the specified image is.

4. The code change doesn't indicate to a user that fragment identifiers are available and which identifier to use. They would need to know how the app is built, by looking at the repo or inspecting the DOM.

I would like to suggest that you open a feature request issue so that we can gauge how much the community needs to jump to an image in Mascots page. Would this be okay with you?

MinThaMie (approved):
Amazing :)

PR #806, opened 2021-03-17
Make Mascots images responsive

simonihmig:
Following up on #769 this makes the mascots page also use responsive images.

### DIFF
--- a/app/styles/components/homepage-image-grid.css
+++ b/app/styles/components/homepage-image-grid.css
@@ -45,6 +45,10 @@
   width: max-content;
 }
 
+[class*=well-] img {
+  width: auto;
+}
+
 /* IN ht estyleguides */
 
 img {
--- a/app/components/mascots/mascot-list/item.hbs
+++ b/app/components/mascots/mascot-list/item.hbs
@@ -1,12 +1,11 @@
 <figure>
   <div class="well well-1/1">
-    <img
+    <ResponsiveImage
+      @src="/images/tomsters/{{@mascot.image}}"
       alt=""
-      data-test-field="Image"
-      loading="lazy"
       role="presentation"
-      src="/images/tomsters/{{@mascot.image}}"
-    >
+      data-test-field="Image"
+      />
   </div>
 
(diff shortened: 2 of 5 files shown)


simonihmig:
I put this PR into draft mode, as I was observing some strange layouts for the mascots page on mobile, and I initially thought this was related to my changes. However it was not, see #807 

Nevertheless this is blocking this effort here, because I need to know how the layout actually should be, as this affects the optimal image sizes.

ijlee2:
Thanks, @simonihmig. I asked Nick to have a look at https://github.com/ember-learn/ember-styleguide/pull/356. Let's wait for a response.

simonihmig:
Ok, I will revisit this, after the blocking issue seems to be resolved.

app/styles/components/homepage-image-grid.css
```diff
@@ -45,6 +45,10 @@
   width: max-content;
 }
 
+[class*=well-] img {
+  width: auto;
```
simonihmig, line 49: This was needed as the styles for images in wells (coming from ember-styleguide) would cause images with defined `width` and `height` attributes (which ember-responsive-images sets automatically and is a [recommended practice again](https://www.smashingmagazine.com/2020/03/setting-height-width-images-important-again/)) to have distorted dimensions (ignoring the explicit aspect ratio given by these attributes) in Safari and FF (but not Chrome!). 

The easiest fix for this was to add this here, but I can also try to upstream this (which should not have any adverse effects when the attributes are not set) to ember-styleguide!?

simonihmig (commented):
@ijlee2 This is good for review now. Idk why percy fails to run in CI here though, but seems unrelated as this was working for previous commits... 🤔

ijlee2 (approved):
LGTM, thank you for working on improving the Mascots page!

With the Netlify preview deploy, I tested that the images load on Chrome, Edge, Firefox, and Safari and quickly checked that the filters worked. The failed CI (Percy didn't start) seemed to be a one-off incidence. I reran the CI and it passed.

Issue #807, opened 2021-03-17
[BUG] Inconsistent grid behavior in prod vs. dev


simonihmig:
**Describe the bug**

The grid classes seem to be applied differently when running `ember serve` vs. what you see on emberjs.com. Observed this on the mascots page, but maybe (probably!?) others are involved as well.

**To Reproduce**
1. Force your browser into a mobile layout (e.g. Dev Tools iPhone device)
2. Go to http://localhost:4200/mascots
3. See a 2-column layout
4. Go to https://emberjs.com/mascots/
5. See a one-column layout

**Screenshots**

On emberjs.com:

![image](https://user-images.githubusercontent.com/1325249/111547273-2e268100-8779-11eb-97b6-2970d2d7c8e4.png)

On localhost:

![image](https://user-images.githubusercontent.com/1325249/111547336-4a2a2280-8779-11eb-8343-9004a3650944.png)

See the different order of CSS selectors being applied to the `ul.grid` element, especially with regards to the `--column-count` CSS var.

**Additional context**

https://github.com/ember-learn/ember-styleguide/pull/356 seems to be related /cc @nickschot

simonihmig:
Same for the ember-users page!

Emberjs.com:

![image](https://user-images.githubusercontent.com/1325249/111548539-27007280-877b-11eb-868b-7b57127b4821.png)

localhost:

![image](https://user-images.githubusercontent.com/1325249/111548562-31227100-877b-11eb-9c65-5229adb13945.png)

mansona:
This seems like a very strange bug 🙈 I has reminded me to add something that should really be added to the website for correctness sake https://github.com/ember-learn/ember-website/pull/809 and I'm wondering if this will fix the bug too 🤔 

Is it really only showing a difference in the columns if you have a trailing slash or not? 🙃

ijlee2:
Although #809 has been merged, the style issue is still present in `localhost:4200`.

mansona:
@ijlee2 can you give steps for reproduction? I only ask because the steps in the main description of this issue say that you go to `http://localhost:4200/mascots` and then `http://localhost:4200/mascots/` but now that #809 is merged those two should end up at the same route 🤔

ijlee2:
Hi, @mansona. Both URLs do end up in http://localhost:4200/mascots/.

However, locally, the app still seems to show 2 columns on mobile width. In production, https://emberjs.com/mascots/ correctly shows 1 column on mobile width.

[...]

PR #809, opened 2021-03-18
implement trailing-history location type

mansona:
This implements a trick that we use in all Empress projects to make sure that the url in the browser always matches what you would expect when you navigate to page on the website. It's a little bit hard to explain, and the ramifications are hard to quantify (and vast) but I will be able to explain with a simple statement and an example: 

After merging this PR, when you are navigating around the website you will **always** see a trailing `/` at the end of the URL. 

For example: Right now if you visit a sub-page of the website such as the logos page you will always see a trailing slash regardless if you put it in the url bar. Another way of expressing it is that if you visit https://emberjs.com/logos or https://emberjs.com/logos/ the browser will always add the trailing slash. 

In the current production website this is **not** true if you arrive at the logos page from an internal Ember navigation. e.g. First load https://emberjs.com and then click `About > Branding` in the Navbar and you will see that the browser has https://emberjs.com/logos **with no trailing slash** in the url bar. If you then refresh the browser will add the trailing slash again. 

This PR fixes that silliness 😄

### DIFF
--- a/app/locations/trailing-history.js
+++ b/app/locations/trailing-history.js
@@ -1,0 +1,13 @@
+import HistoryLocation from '@ember/routing/history-location';
+
+export default class TrailingHistoryLocation extends HistoryLocation {
+  formatURL() {
+    let url = super.formatURL(...arguments);
+
+    if (url.includes('#')) {
+      return url.replace(/([^/])#(.*)/, '$1/#$2');
+    } else {
+      return url.replace(/\/?$/, '/');
+    }
+  }
+}
--- a/config/environment.js
+++ b/config/environment.js
@@ -5,7 +5,8 @@ module.exports = function (environment) {
     modulePrefix: 'ember-website',
     environment,
     rootURL: '/',
-    locationType: 'auto',
+    locationType: 'trailing-history',
+    historySupportMiddleware: true,
     EmberENV: {
       FEATURES: {
         // Here you can enable experimental features on an ember canary build


amyrlam (approved):
👍 sgtm

app/locations/trailing-history.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default class TrailingHistoryLocation extends HistoryLocation {
+  formatURL() {
+    let url = super.formatURL(...arguments);
+
+    if (url.includes('#')) {
+      return url.replace(/([^/])#(.*)/, '$1/#$2');
+    } else {
+      return url.replace(/\/?$/, '/');
+    }
```
ijlee2, line 11: Let's make it clear to others that there's an early exit.

```suggestion
    if (url.includes('#')) {
      return url.replace(/([^/])#(.*)/, '$1/#$2');
    }

    return url.replace(/\/?$/, '/');
```

app/locations/trailing-history.js
```diff
@@ -0,0 +1,13 @@
+import HistoryLocation from '@ember/routing/history-location';
+
+export default class TrailingHistoryLocation extends HistoryLocation {
+  formatURL() {
+    let url = super.formatURL(...arguments);
```
ijlee2, line 5: What is the type of `url`, returned from `formatURL`? Is it guaranteed to be a string?

ijlee2 (changes requested):
Before we merge the pull request, I'd like to see unit tests added. I think the regular expressions are not easy to follow and will discourage people from modifying the `trailing-history` file when needed.

Using the following terms,

```
URI = scheme ":" hierarchical-part ["?" query] ["#" fragment]
```

can you add a unit test file with 2 submodules and 4 cases (a total of 8 tests), to describe the intended outputs for:

- Submodule 1. `When the URI's hierarchical part doesn't end in a forward slash`
- Submodule 2. `When the URI's hierarchical part ends in a forward slash`

The 4 cases are:

- Case 1. `When there are no query parameters and no fragment identifier`
- Case 2. `When there are only query parameters`
- Case 3. `When there is only a fragment identifier`
- Case 4. `When there are query parameters and a fragment identifier`

config/environment.js
```diff
@@ -5,7 +5,8 @@ module.exports = function (environment) {
     modulePrefix: 'ember-website',
     environment,
     rootURL: '/',
-    locationType: 'auto',
+    locationType: 'trailing-history',
+    historySupportMiddleware: true,
```
ijlee2, line 9: For more information, I think we can look at https://dockyard.com/blog/2019/12/12/ember-location-api-part-2-extending-and-customizing

app/locations/trailing-history.js
mansona: but there isn't an "early exit" 🤔  both paths do different things but they are both doing something to the URL so I would be uncomfortable making this look like an early exit

app/locations/trailing-history.js
mansona: This will always be a string for sure, unless we have a serious problem in the whole use of URLs in every ember app ever made 😂

amyrlam:
@ijlee2 we talked about this at the learning team meeting with @locks @jaredgalanis 

@mansona confirmed this is already in the guides production, from a couple years ago, so it doesn't seem necessary to test in this repo right now. we decided it would be better to merge it in to partiallysolve the bug https://github.com/ember-learn/ember-website/issues/807 

there was some discussion about locationType: auto might go away

we weren't sure what you meant by: https://github.com/ember-learn/ember-website/pull/809#discussion_r598452657

we could talk about it at a future meeting also and/or on discord, potentially make a follow-on issue, to be sure your questions / comments are addressed?

Issue #794, opened 2021-03-07
Use VerticalBarChart to create 2017 survey responses


ijlee2:
## Background

Until recently, the controller files for `ember-community-survey` pages included hundreds of hard-coded configuration objects for Highcharts. These objects were large and only worked with Highcharts. As a result, the objects were difficult to review quickly and posed a technical risk in maintenance.

To solve these issues, I wrote classes that abstract creating a configuration object. These classes accept a minimal set of arguments so that we can better adapt to a breaking change in a Highcharts release or a migration to another chart library when needed.

## Description

In #793, I refactored the `ember-community-survey-2020` controller by:

- Finding configuration objects that have `type: 'column'`
- Rewriting such objects as instances of the `VerticalBarChart` class
- Rewriting percentage values in the form of `100 * (count / total)`

To complete the issue, please refactor the `ember-community-survey-2017` controller in a similar manner. (I believe you will get to **convert 10 objects**.)

## Screenshots

I added a screenshot to help provide additional context. The `VerticalBarChart` class is used to create bar charts that look like the following:

<img width="640" alt="a vertical bar chart" src="https://user-images.githubusercontent.com/16869656/110236090-3a6a3d00-7f34-11eb-91b2-3504c2189ad2.png">

The image shows 4 **series** of data: ARIA, CSS, HTML, and JavaScript. Each series has data that belong to 3 **categories**: Beginner, Intermediate, and Advanced.

fozy81:
Hi @ijlee2 , I'd like to give this a go.

ijlee2:
Hi, @fozy81. Thank you for helping out! Let me know if you have questions.

PR #810, opened 2021-03-22
fix: issue-794 - update column charts to VerticalBarChart class

fozy81:
See #794 issue


fozy81:
Hi @ijlee2 could you review this change? I've not changed the percentage to '100 * (count / total)' as requested, is it necessary?

app/controllers/ember-community-survey-2017.js
```diff
@@ -531,232 +451,195 @@ const recommendingEmber = new VerticalBarChart({
...
-    },
-    xAxis: {
-      type: 'category',
-    },
-    yAxis: {
-      title: {
-        text: 'Percent',
-      },
-    },
-    drilldown: {
```
ijlee2, line 705: Good catch. This is the first time I'm encountering `drilldown` property.

I checked that, in the production site, the drilldown doesn't work (clicking on North America bar did nothing) and doesn't also error.

Since this is the only place is `drilldown` is used (the `drilldown` data in 2018 is a copy-paste of that in 2017), I'll remove it in a separate pull request.

app/controllers/ember-community-survey-2017.js
```diff
@@ -391,118 +341,88 @@ const longLived = {
...
-        { y: 6, name: 'Elixir / Erlang' },
-        { y: 5, name: 'Go' },
-        { y: 2, name: 'Scala' },
-        { y: 0, name: 'TypeScript' },
-        { y: 0, name: 'C++' },
-        { y: 1, name: 'Rust' },
-        { y: 3, name: 'Other' },
-      ],
+      label: '2016',
+      values: [41, 24, 17, 13, 12, 8, 6, 5, 2, 0, 0, 1, 3],
```
ijlee2, line 371: Before the changes, the `data` arrays for 2016 and 2017 didn't list the categories in the same order (e.g. PHP and Python), so we unfortunately want to be careful with creating tuples. Luckily, we have Percy to tell us if we might have switched an order.

```suggestion
      values: [41, 24, 17, 12, 13, 8, 6, 5, 2, 0, 0, 1, 3],
```

app/controllers/ember-community-survey-2017.js
```diff
@@ -391,118 +341,88 @@ const longLived = {
...
+      'Python',
+      'PHP',
+      '.NET',
+      'Elixir / Erlang',
+      'Go',
+      'Scala',
+      'TypeScript',
+      'C++',
+      'Other',
+    ],
```
ijlee2, line 363: Let's use 2016, which has more items, as the source of truth for categories.

```suggestion
    categories: [
      'Ruby',
      'JavaScript',
      'Java',
      'PHP',
      'Python',
      '.NET',
      'Elixir / Erlang',
      'Go',
      'Scala',
      'TypeScript',
      'C++',
      'Rust',
      'Other',
    ],
```

app/controllers/ember-community-survey-2017.js
```diff
@@ -391,118 +341,88 @@ const longLived = {
...
+        20.57,
+        13.5,
+        12.43,
+        8.71,
+        7.95,
+        6.44,
+        2.65,
+        1.58,
+        1.2,
+        2.97,
```
ijlee2, line 388: Before the changes, the `data` arrays for 2016 and 2017 didn't list the categories in the same order (e.g. PHP and Python). 2017 is also missing `Rust`.

As a result, we unfortunately want to be careful when creating tuples. Luckily, we have Percy to tell us if we might have switched an order.

percy snapshot

```suggestion
      values: [
        36.97,
        24.54,
        20.57,
        12.43,
        13.5,
        8.71,
        7.95,
        6.44,
        2.65,
        1.58,
        1.2,
        0,
        2.97,
```

app/controllers/ember-community-survey-2017.js
```diff
@@ -531,232 +453,195 @@ const recommendingEmber = new VerticalBarChart({
...
+//         { name: 'Europe	', y: 33.6, drilldown: 'Europe' },
+//         { name: 'Asia	', y: 7.8, drilldown: 'Asia' },
+//         { name: 'Oceania	', y: 3.7 },
+//         { name: 'South America	', y: 3.5 },
+//         { name: 'Africa	', y: 0.8 },
+//         { name: 'Central America	', y: 0.4 },
+//       ],
+//     },
+//   ],
+// };
```
ijlee2, line 644:
```suggestion
```

ijlee2 (approved):
@fozy81 Thank you for your first contribution to the `ember-website` repo and helping me maintain the controller files.

I thought you did well with making a good judgment call when there were 2 uncertainties regarding percentage format and drilldown property. (I didn't realize there were no percentage examples in 2017, which would lead to confusion. My bad.)

Thanks again for helping out! 🥳

fozy81:
Thanks for reviewing and tidying up the loose ends @ijlee2 :tada:

PR #813, opened 2021-03-22
Update teams.hbs

MelSumner:
Gives each section more breathing room.


amyrlam (approved):
Looks good! Spacing of avatars in a row looked odd in the prior PR when removed one person from steering and were only 4...but since you added one back it's back to 5 now, should be g2g

PR #817, opened 2021-03-23
Removed drilldown (does not work)

ijlee2:
## Description

This is a follow-up PR to #810. We discovered that there was 1 use of Highcharts' `drilldown` feature, but the feature did not work in the production site. This PR cleans up remaining code.

### DIFF
--- a/app/controllers/ember-community-survey-2018.js
+++ b/app/controllers/ember-community-survey-2018.js
@@ -627,43 +627,6 @@ const demographicsRegion = {
     subtitle: { text: '' },
     xAxis: { type: 'category' },
     yAxis: { title: { text: 'Percent' } },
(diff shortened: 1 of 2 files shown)


app/controllers/ember-community-survey-2018.js
```diff
@@ -627,43 +627,6 @@ const demographicsRegion = {
...
-        name: 'Asia',
-        id: 'Asia',
-        data: [
-          {name: 'Asian', y: 67},
-          {name: 'White', y: 12},
-          {name: 'Prefer not to answer', y: 8.9}
-        ]
-      }]
-    }
-    */
```
ijlee2, line 666: Copy-paste from 2017

PR #834, opened 2021-04-03
Installed ember-exam (trial)

ijlee2:
## Description

I tried to install `ember-exam` to speed up continuous integration. I'm closing the pull request because the CI is currently turning out to be slower. Here are a few discoveries:

1\. It takes too long to build a test app and upload it for caching. We need to identify what in `ember-website` project causes these to happen.

2\. We're building the production app 2 times (for lighthouse and asset sizes). Both processes need to be faster, and we need to look into building a production app once.

3\. Each time a package is installed with `npm`, Netlify seems to be unable to detect a change and we need to manually clear the cache and redeploy the preview site. Look into the underlying issue.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,7 +9,7 @@ on:
 env:
   NODE_VERSION: 14
   PERCY_PARALLEL_NONCE: ${{ github.run_id }}-${{ github.run_number }}
-  PERCY_PARALLEL_TOTAL: 1
+  PERCY_PARALLEL_TOTAL: 4
 
 jobs:
   lint:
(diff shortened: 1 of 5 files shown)


mansona:
I'm not sure we need the complexity of ember-exam for the website at all 🤔  Yes we build the site multiple times but the thing that takes most of the time is the actual Lighthouse run, which runs multiple pages multiple times to get an average of the lighthouse score. Everything else runs in parallel so it's not being held up, it's just the critical path. 

And looking at one of the latest builds it seems like the longest job (the lighthouse build) only takes just over 3 minutes to finish https://github.com/ember-learn/ember-website/actions/runs/722404528 . If our CI was taking something close to 10 minutes then I think we would need some sort of intervention.

I also had a few conversations with @Turbo87 and I'm not too sure how much time we're saving by implementing quite a complex caching strategy, we might be worth trying the build without it and see if it makes a tangible difference.

And as for the issues with Netlify, I've contacted them a few times about this and I have requested that they use `npm ci` to install dependencies so that it makes sure all dependencies always match the package-lock but they aren't open to this change 😞  on top of that it seems that they are doing something custom with their caching strategy that can sometimes be incompatible with npm (or yarn) so we will sometimes have this issue where you need to force a cache clear and rebuild if it fails.

PR #836, opened 2021-04-03
Update logos.hbs

esquith:


ijlee2:
I will close the pull request because both all right and alright are acceptable spellings.

PR #838, opened 2021-04-05
Refactor hardcoded values into SplineChart

### LINKED ISSUE #826: 2021 Contributors' Workshop: Create SplineChart class
## Background

Until recently, the controller files for `ember-community-survey` pages included hundreds of hard-coded configuration objects for Highcharts. These objects were large and only worked with Highcharts. As a result, the objects were difficult to review quickly and posed a technical risk in maintenance.

To solve these issues, I have been writing classes that abstract creating a configuration object. These classes accept a minimal set of arguments so that we can better adapt to a breaking change in a Highcharts release or a migration to another chart library when needed.

## Description

First, please have a look at #784 and #778 (their first 4 commits), to see how I had created a class and wrote unit tests in a step-by-step manner.

To complete the issue, please create a class called `SplineChart` in `app/utils/highcharts` folder. The `SplineChart` class can be used to create configuration objects for chart type of `'spline'` in `ember-community-survey-2017` and `ember-community-survey-2018`.

Don't forget to write unit tests for `SplineChart`. 🙂

## Notes

The class is to adhere to the API of other, similar classes in this folder. In other words, the class accepts an object with 2 properties called `chart` and `rawData`. As output, the `highchartsOptions` returns the configuration object for rendering a spline chart.

To limit the scope of work and PR review, I recommend applying the `SplineChart` class to replace existing configuration objects in a separate issue and pull request.

mamiller93:
## Description
A continuation of #784, where @ijlee2 started refactoring the code in `ember-community-survey` controller files. In this pull request, I created the `SplineChart` class, which can replace 4 instances of hard-coded chart options. 
 
## Need help
Right now the tests don't pass because the function signature (I assume) are different between the test input and expectation, despite the serialization being equivalent and using a `deepEqual` assertion. I am passing a function so the `SpineChart` doesn't need to dictate how the tooltip function should behave. I'm open to different ideas!

Closes #826


mamiller93:
@ijlee2 could you take a look and give feedback as to my "needs help" section above? I'm not sure how to proceed with qunit as I'm not exactly sure what is really the root cause issue here.

tests/unit/utils/highcharts/spline-chart-test.js
```diff
@@ -84,9 +88,6 @@ module('Unit | Utility | highcharts/spline-chart', function () {
 
           tooltip: {
             crosshairs: true,
-            formatter: function tooltip() {
-              return `${this.point}`;
-            },
```
ijlee2: I imagine that you might have had to pass the same function in order for `deepEqual` to pass, similarly to how, in JavaScript, `{} === {}` is not equal to `true` because the objects refer to a different instance.

```javascript
const formatterFunction = ...;

// Setup
tooltip: formatterFunction

// Assertion
formatter: formatterFunction
```

Let's avoid checking the formatter function by deleting it from `options`.

ijlee2 (approved):
I think there's more refactoring/simplification that we can do for the spline charts (with regards to categories and tooltip formatter). Nonetheless, your pull request, which focused on creating the utility, makes a great start.

Thanks very much for helping out and making your first contribution to the `ember-website` repo!

PR #842, opened 2021-04-15
Updated SplineChart class

ijlee2:
## Description

This pull request continues work from #838.

Until now, we rendered 4 highly customized spline charts in 2017 and 2018 survey results. These charts superimposed graphs in an attempt to indicate how a trend might have changed over time.

what spline chart class can render currently

The problem with this approach is the high customization. We overrode the default tooltip with a complex, untested function and could not reuse the spline chart class to visualize any other data application.

A simpler approach can reduce the bundle size (by ~1.2 KB, gzipped), allow us to reuse the class for other data applications, and increase maintainbility. In addition, the new graph can help a general public understand the message that we want to tell with historical data:

what spline chart class will render from now on

### DIFF
--- a/app/templates/ember-community-survey-2017.hbs
+++ b/app/templates/ember-community-survey-2017.hbs
@@ -17,19 +17,10 @@
   @sectionTitle="Ember Version"
 >
   <:charts>
-    <div>
-      <Highcharts
-        @chartOptions={{this.chartHowlong.options}}
-        @data={{this.chartHowlong.data}}
-      />
-    </div>
-
-    <div class="mt-3">
-      <Highcharts
-        @chartOptions={{this.chartBreakdownByVersion.options}}
-        @data={{this.chartBreakdownByVersion.data}}
-      />
-    </div>
+    <Highcharts
+      @chartOptions={{this.chartHowlong.options}}
+      @data={{this.chartHowlong.data}}
+    />
   </:charts>
 
   <:body>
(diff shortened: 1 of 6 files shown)


app/templates/ember-community-survey-2017.hbs
```diff
@@ -17,19 +17,10 @@
...
-        @chartOptions={{this.chartHowlong.options}}
-        @data={{this.chartHowlong.data}}
-      />
-    </div>
-
-    <div class="mt-3">
-      <Highcharts
-        @chartOptions={{this.chartBreakdownByVersion.options}}
-        @data={{this.chartBreakdownByVersion.data}}
-      />
```
ijlee2, line 31: I removed the second chart because the new spline chart (`chartHowlong`) displays the same data as the bar chart (`chartBreakdownByVersion`), but is easier to read and contains one additional year of data.

app/templates/ember-community-survey-2018.hbs
```diff
@@ -17,19 +17,10 @@
...
-        @chartOptions={{this.newReleaseAdoption.options}}
-        @data={{this.newReleaseAdoption.data}}
-      />
-    </div>
-
-    <div class="mt-3">
-      <Highcharts
-        @chartOptions={{this.overallAdoption.options}}
-        @data={{this.overallAdoption.data}}
-      />
```
ijlee2, line 31: I removed the second chart because the new spline chart (`newReleaseAdoption`) displays the same data as the bar chart (`overallAdoption`), but is easier to read and contains one additional year of data.

PR #846, opened 2021-04-24
Remove unused Highcharts code (Part 1)

### LINKED ISSUE #845: Remove unused Highcharts code (Part 1)
## Background

Until recently, all Highcharts objects were manually configured, then _merged_ with a default configuration object. 

Now that configuration objects are created as an instance of a chart class, we can remove code that had been based on the prior approach.

## Description

We'll remove unused code in a few steps (parts). The goal of the 1st step is to **simplify the API of `<Highcharts>` component**.

### 1. Current approach

Once we create an instance of a chart class, we have to pass the instance's `options` and `data` properties separately. This architectural approach allowed backwards compatibility.

As an illustration, in an `ember-community-survey` controller, suppose that we have `chartHowLong`:

```javascript
const chartHowlong = new VerticalBarChart( ... ).highchartsOptions;
```

Note that `chartHowlong` is an **instance property** of the `VerticalBarChart` class.

In the corresponding route template, we would write,

```handlebars
<Highcharts
  @chartOptions={{this.chartHowlong.options}}
  @data={{this.chartHowlong.data}}
/>
```

### 2. New approach

Recall that the goal of this issue is to simplify the API of the `<Highcharts>` component.

#### Step 1

Wherever the component is invoked (in the `ember-community-survey` controllers), we want to pass just 1 argument, called `@chart`:

```handlebars
{{!-- Before --}}
<Highcharts
  @chartOptions={{this.chartHowlong.options}}
  @data={{this.chartHowlong.data}}
/>

{{!-- After --}}
<Highcharts
  @chart={{this.chartHowlong}}
/>
```

To make a minimal change to the component's class, please update only the getter `chartOptions`.

https://github.com/ember-learn/ember-website/blob/19b5e447f7465757dfd6c84fcbe1ab69f1caaefd/app/components/highcharts.js#L18-L22

Afterwards, please check that all tests pass. (You may need to update some tests.)

#### Step 2

Next, we will redefine `chartHowlong`. Before, it represented an instance property of a chart. Now, it will represent the instance itself (the "whole object," if you will):

```javascript
// Before
const chartHowlong = new VerticalBarChart( ... ).highchartsOptions;

// After
const chartHowlong = new VerticalBarChart( ... );
```

Note that we want to redefine `chartHowLong` so that the following template code makes sense:

```handlebars
<Highcharts
  @chart={{this.chartHowlong}}
/>
```

That is, both left-hand side (`@chart`) and right-hand side (`this.chartHowlong`) represent a chart.

Once again, you will need to update the getter `chartOptions` in `<Highcharts>`'s component class. Please check that all tests pass.

srih4ri:
- Update highcharts component to accept instance of a chart as @chart
- Update usage of highcharts component in templates
- Update controllers to pass instance of Chart object

@ijlee2 I have handled not receiving a '@chart' argument to support the current tests for 'ember-community-survey/section' - we don't pass a '@chart' argument here. Not sure if this is the right way to go.

Closes #845

### DIFF
--- a/app/components/highcharts.js
+++ b/app/components/highcharts.js
@@ -16,7 +16,8 @@ export default class HighchartsComponent extends Component {
   @tracked isChartDrawn = false;
 
   get chartOptions() {
-    const { chartOptions, data } = this.args;
+    const { options: chartOptions, data } =
+      this.args.chart?.highchartsOptions || {};
 
     return createChartOptions({ chartOptions, data });
   }
--- a/app/components/highcharts.js
+++ b/app/components/highcharts.js
@@ -16,8 +16,13 @@ export default class HighchartsComponent extends Component {
   @tracked isChartDrawn = false;
 
   get chartOptions() {
-    const { options: chartOptions, data } =
-      this.args.chart?.highchartsOptions || {};
+    const { chart } = this.args;
+
+    if (!chart) {
+      return {};
+    }
+
+    const { options: chartOptions, data } = chart.highchartsOptions;
 
     return createChartOptions({ chartOptions, data });
   }
--- a/app/controllers/ember-community-survey-2016.js
+++ b/app/controllers/ember-community-survey-2016.js
@@ -29,7 +29,7 @@ const chartHowlong = new VerticalBarChart({
       values: [27, 16, 16, 9, 9, 10, 10],
     },
   ],
-}).highchartsOptions;
+});
 
 const chartEmberAtTheOffice = [
   new VerticalBarChart({
@@ -56,7 +56,7 @@ const chartEmberAtTheOffice = [
         values: [59, 15, 16, 11, 23],
       },
     ],
-  }).highchartsOptions,
+  }),
 
   new VerticalBarChart({
     chart: {
@@ -76,7 +76,7 @@ const chartEmberAtTheOffice = [
         values: [12, 74, 6, 4, 2, 1],
       },
     ],
-  }).highchartsOptions,
+  }),
 ];
 
 const chartNewReleaseAdoptionEmber = new VerticalBarChart({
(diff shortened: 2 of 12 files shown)


app/components/highcharts.js
```diff
@@ -16,7 +16,8 @@ export default class HighchartsComponent extends Component {
   @tracked isChartDrawn = false;
 
   get chartOptions() {
-    const { chartOptions, data } = this.args;
+    const { options: chartOptions, data } =
+      this.args.chart?.highchartsOptions || {};
```
ijlee2, line 20: As an aside, when you want to provide a default value, I recommend using the [nullish coalescing operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator), which has the syntax of `??`.

The OR operator, `||`, allows the 2nd operand to take when the 1st operand is falsey. On the other hand, the `??` allows the 2nd operand only when the 1st operand is either `undefined` or `null`.

I think having limited, very specific choices makes it less likely to introduce bugs and makes your intent (that the 2nd operand, `{}`, is the default value) more clear.

ijlee2 (approved):
@srih4ri Thanks again for helping out and wanting to practice more Ember.

You did great with updating the controllers and templates, and showed understanding of how line 19 in `app/components/highcharts.js` would need to change, especially when there's a constraint set by existing tests.

I pushed a couple of extra commits in preparation of Part 2. The first extra commit shows, what one might do to balance code readability and getting existing tests to pass. The second commit, I added a sample chart instance that rendering tests for `<Highcharts>` can use. I thought this could also help improve documentation for `<Highcharts>`.

Can you study the code changes in the last 2 commits? I will set up Part 2 and let you know when it's ready.

app/components/highcharts.js
```diff
@@ -16,8 +16,13 @@ export default class HighchartsComponent extends Component {
...
   get chartOptions() {
-    const { options: chartOptions, data } =
-      this.args.chart?.highchartsOptions || {};
+    const { chart } = this.args;
+
+    if (!chart) {
+      return {};
+    }
+
+    const { options: chartOptions, data } = chart.highchartsOptions;
```
ijlee2, line 25: We'll clean up lines 25-27 in Part 2.

PR #857, opened 2021-07-19
Make ember-users images responsive

simonihmig:
It required adding a conditional using a custom helper to check for SVGs used as the company logo (by a handful of companies), in which case responsive images using `<picture>` and `srcset` do not make sense.

*Actually*, ideally all company logos should be SVGs IMO. But we cannot change that here without requiring all users to upload new logo files...

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -103,6 +103,18 @@ module.exports = function (defaults) {
           // don't scale images, just copy as-is in dev mode, to not slow down the build
           justCopy: process.env.EMBER_ENV !== 'production',
         },
+        {
+          include: 'images/users/*.png',
+          widths: [
+            58, // mobile
+            116, // mobile 2x
+            88, // desktop
+            176, // desktop 2x, iPhone 2x
+          ],
+          removeSource: true,
+          // don't scale images, just copy as-is in dev mode, to not slow down the build
+          justCopy: process.env.EMBER_ENV !== 'production',
+        },
       ],
     },
   });
--- a/app/helpers/file-extension.js
+++ b/app/helpers/file-extension.js
@@ -1,0 +1,6 @@
+import { helper } from '@ember/component/helper';
+
+export default helper(function fileExtension([filename]) {
+  // taken from https://stackoverflow.com/a/12900504 for optimal performance
+  return filename.slice(((filename.lastIndexOf('.') - 1) >>> 0) + 2);
+});
--- a/app/templates/ember-users.hbs
+++ b/app/templates/ember-users.hbs
@@ -25,11 +25,18 @@
       }}
         <li>
           <a href={{user.url}} rel="nofollow noopener noreferrer" target="_blank" class="well well-16/9">
-            <img
-              alt={{user.name}}
-              src="/images/users/{{user.image}}"
-              loading="lazy"
-            >
+            {{#if (eq (file-extension user.image) "svg")}}
+              <img
+                alt={{user.name}}
+                src="/images/users/{{user.image}}"
+                loading="lazy"
+              >
+            {{else}}
+              <ResponsiveImage
+                @src="/images/users/{{user.image}}"
+                alt={{user.name}}
+              />
+            {{/if}}
           </a>
         </li>
       {{/each}}
--- a/tests/integration/helpers/file-extension-test.js
+++ b/tests/integration/helpers/file-extension-test.js
@@ -1,0 +1,20 @@
+import { module, test } from 'qunit';
+import { setupRenderingTest } from 'ember-qunit';
+import { render } from '@ember/test-helpers';
+import { hbs } from 'ember-cli-htmlbars';
+
+module('Integration | Helper | file-extension', function (hooks) {
+  setupRenderingTest(hooks);
+
+  test('it extracts the file extension', async function (assert) {
+    await render(hbs`
+      <div data-test-value>
+        {{file-extension "some.file.png"}}
+      </div>
+    `);
+
+    assert
+      .dom('[data-test-value]')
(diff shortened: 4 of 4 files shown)


simonihmig:
Put this PR into draft mode, as the percy CI run revealed that the user images were also used on the sponsors page. Will address that tomorrow...

simonihmig:
I have updated this PR to also include the sponsors images. To be able to better separate users and sponsors images - the latter are required in different (larger) resolutions - I copied/moved the images used on the sponsors page into a separate folder. In same cases this creates a little bit of duplication, but otherwise I would have to generate the larger sizes for *all* user images, which wouldn't even be used.

**There is one interesting issue though**. For some reason I don't fully understand, the responsive images (using  the advanced markup with `<picture>` and so on) somehow behave differently in the specific scenario of images placed in `.well`. The "unusal" part here is that the images don't have `width` set as a concrete value (either `px` or `%` or so) in CSS (it's `auto`), instead the width is defined implicitly through `max-width` and `max-height`. So the issue is that *now* (again, I don't fully understand why the new markup causes a change in behavior) images that are not quite large enough to fill the maximum size  without getting blurry (e.g. an image that needs to be 200px on a retina display should be 400px wide) will be smaller than before.

This is the sponsors page on my MBP (retina display!):

image

And this is the same page, when I reload it after I have moved the tab to my 2nd monitor (non-retina):

image

The latter should match exactly what we have now (percy is happy!). As you can see the images that are large enough (e.g. "201 created") have the same size, while all logos that are just 210px wide appear too small.

So this is quite weird but interesting. Probably this is correct behavior by the browser according to the specs, but I am just not aware what the rules are there. Maybe some better CSS expert can explain?

For the solution, I tried to change this behavior, but failed. On the other hand, scaling the image to the full possible size on a retina display does make them blurry, so the solution with the best outcome in quality would be to replace the existing images that are too small with larger ones if possible. They should be at least 468px, but can be larger as the addon will rescale them. This way they would have the correct size *and* be crystal clear on screens with high pixel density. But this probably requires contacting the (past!) sponsors to deliver larger ones, which is beyond my authority. 

Or live with the way it is now, which is not bad at all IMHO, just not perfect.

🤷‍♂️

simonihmig:
As a bit of motivation to land this, when comparing the image weight of the users page when all images are (lazily) loaded:

Before: 3.0MB
After: 1.9MB

Also, when doing a full reload of the users page, for the existing implementation the images are *not* lazily loaded, despite having `loading="lazy"`. You can see this in the network tab, all are loaded eagerly. With the addon this is fixed. This might be caused by similar issues related to how GlimmerVM dynamically created the DOM elements that I faced in https://github.com/kaliber5/ember-responsive-image/pull/176. 🤔

app/templates/sponsors.hbs
```diff
@@ -21,7 +21,19 @@
...
+              {{#if (eq (file-extension sponsor.image) "svg")}}
+                <img
+                  alt="{{sponsor.name}} website"
+                  src="/images/sponsors/{{sponsor.image}}"
+                  loading="lazy"
+                >
+              {{else}}
+                <ResponsiveImage
+                  @src="/images/sponsors/{{sponsor.image}}"
+                  @sizes="(min-width: 1008px) 234px, (min-width: 375px) 90px, 60px"
```
mansona, line 33:
```suggestion
                  @sizes="(min-width: 1008px) 234px, (min-width: 375px) 180px 90px, 60px"
```

jenweber:
Hi @simonihmig thanks for your work on this and the iterations! We reviewed this after the learning team meeting, and we noticed that the proportions seem off for ipad sized screens. @mansona will follow up with a code suggestion. Otherwise, seems great!

app/templates/sponsors.hbs
```diff
@@ -49,7 +61,19 @@
...
```
mansona, line 73:
```suggestion
                  @sizes="(min-width: 1008px) 234px, (min-width: 375px) 180px 90px, 60px"
```

ember-cli-build.js
```diff
@@ -103,6 +103,31 @@ module.exports = function (defaults) {
...
           justCopy: process.env.EMBER_ENV !== 'production',
         },
+        {
+          include: 'images/users/*.png',
+          widths: [
+            60, // mobile
+            120, // mobile 2x
+            90, // desktop
+            180, // desktop 2x, iPhone 2x
+          ],
```
mansona, line 113:
```suggestion
          widths: [
            60, // mobile
            120, // mobile 2x
            90, // desktop
            180, // desktop 2x, iPhone 2x
            360, 
          ],
```

mansona (commented):
This is the suggestion that @jenweber was talking about. I don't really know if this is the right way to achieve what we're looking for here but maybe you can give us a bit of guidance?

app/templates/sponsors.hbs
simonihmig: I don't think this is valid syntax, is it? See https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/sizes

And what exactly are you trying to achieve with this?

Tbh I haven't understood what exactly the problem is? I am aware of the images being too small, as I explained above. What else have you seen that needs fixing? Maybe post a screenshot?

mansona: Yes sorry I don't actually know what the right syntax is 🙈  

Here is a set of images that look fine: 

Large: 
deploy-preview-857--ember-website netlify app_ember-users_

Smallest: 
deploy-preview-857--ember-website netlify app_ember-users_ (1)

But here is the one that now looks different and the images are super small: 
deploy-preview-857--ember-website netlify app_ember-users_ (3)

PR #868, opened 2021-10-04
Fixed incorrect migration to @percy/cli

ijlee2:
## Description

In #867, I hadn't updated the `test` command in `package.json`. Because the `test` command had already included the words `percy exec`, Percy didn't run in CI when the pull request had been merged.

Screenshot 2021-10-04 at 19 29 38


.github/workflows/ci.yml
```diff
@@ -87,7 +87,7 @@ jobs:
         if: steps.cache-dependencies.outputs.cache-hit != 'true'
 
       - name: Test
-        run: npx percy exec -- yarn test
+        run: npx percy exec -- npm run test
```
rwwagner90, line 90: Is there a reason this was changed from yarn?

ijlee2: Yeah. `ember-website` uses `npm` instead of `yarn` to manage packages. I had initially written `yarn test` because I had copy-pasted the solution from `ember-octane-vs-classic-cheat-sheet`.

Issue #862, opened 2021-08-17
[BUG] Filtering in /mascots is not working


ijlee2:
**Describe the bug**

When a user visits https://emberjs.com/mascots, then clicks on one of the filters (e.g. `Meetups`), the user does not see _any_ Meetup-related mascots.

I think a URL malformation is the issue. For some reason, the query parameter `filter` includes a space `%2F` at the end, e.g. `https://emberjs.com/mascots?filter=meetup%2F/`.

When the user clicks on `Meetups`, then on `Small`, we can see `%2F%2F` (i.e. more than one of `%2F`).

We do have application tests that check for the query parameter. In short, how the URL is formed may be different between production and testing.

- [A user can filter mascots](https://github.com/ember-learn/ember-website/blob/main/tests/acceptance/mascots-test.js#L51-L63)
- [A user can change the size of mascots](https://github.com/ember-learn/ember-website/blob/main/tests/acceptance/mascots-test.js#L95-L108)

**Expected behavior**

When the user clicks on a filter, the Mascots page will be able to limit which mascots are shown to the user.

When the user refreshes the page (with the `filter` query parameter on), the Mascots page will be able to limit which mascots are shown to the user.

**Recording**

![recording](https://user-images.githubusercontent.com/16869656/129674533-6e0b7704-c7be-45fa-9ee2-bdbe91621012.gif)

**Desktop (please complete the following information):**
 - OS: Mac
 - Browser: Chrome
 - Version: 92.0.4515.131

claeusdev:
if nobody has claimed this i would like to @ijlee2

mansona:
Sorry this is as a result of the thing we need to add the trailing slash to the end of the url: https://github.com/ember-learn/ember-website/blob/main/app/locations/trailing-history.js

It seems like I didn't correctly support query parameters in my implementation 🤔

ijlee2:
Hi, @claeusdev. Thank you for volunteering to help. I'll go ahead and assign the issue to you.

@rajakvk: To confirm, were you looking into solving the issue as well?

PR #869, opened 2021-10-04
fix issue-862-mascots

fozy81:
Fixing https://github.com/ember-learn/ember-website/issues/862 - mascots filtering

This is alternative to https://github.com/ember-learn/ember-website/pull/863

The fix is a bit hacky...

I think it works but is there a better solution?

### DIFF
--- a/tests/unit/utils/format-url-test.js
+++ b/tests/unit/utils/format-url-test.js
@@ -1,0 +1,24 @@
+import { formatURLConditional } from 'ember-website/utils/format-url';
+import { module, test } from 'qunit';
+
+module('Unit | Utility | format-url', function () {
+  module('formatURLConditional', function () {
+    test('We can format URL', function (assert) {
+      assert.strictEqual(
+        formatURLConditional('https://emberjs.com/logos#Branding'),
+        'https://emberjs.com/logos/#Branding',
+        'If # in URL, we add / before #'
+      );
+      assert.strictEqual(
+        formatURLConditional('http://localhost:4200/mascots?filter=tomster/'),
+        'http://localhost:4200/mascots?filter=tomster',
+        'If ? in URL, we remove trailing /'
+      );
+      assert.strictEqual(
+        formatURLConditional('http://localhost:4200/community'),
+        'http://localhost:4200/community/',
+        'If neither # or ? in URL, we add trailing /'
+      );
+    });
+  });
+});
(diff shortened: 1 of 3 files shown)


ijlee2 (commented):
@fozy81 I'd like to see unit tests added to document the if-else branch. Can you extract the conditional into a function in a utility file, then write at least 3 unit tests (one for every if-branch) to document what `formatURL` is supposed to do?

fozy81:
Thanks for the advice @ijlee2 , I'll write the some tests.

fozy81:
@ijlee2 added tests and extract conditionals into util function. Please take a look and happy to fix anything else required.

tests/unit/utils/format-url-test.js
```diff
@@ -0,0 +1,24 @@
...
+        formatURLConditional('http://localhost:4200/mascots?filter=tomster/'),
+        'http://localhost:4200/mascots?filter=tomster',
+        'If ? in URL, we remove trailing /'
+      );
+      assert.strictEqual(
+        formatURLConditional('http://localhost:4200/community'),
+        'http://localhost:4200/community/',
+        'If neither # or ? in URL, we add trailing /'
+      );
+    });
```
ijlee2, line 22: Thanks for providing realistic URLs as examples. I'd suggest that we create a nested module for each of the 3 cases:

```js
module('Unit | Utility | format-url', function () {
  module('When # is in the URL', function () {
    test('We add / before #', function (assert) {
      const url = formatURLConditional('https://emberjs.com/logos#Branding');

      assert.strictEqual(
        url,
        'https://emberjs.com/logos/#Branding',
        'If # in URL, we add / before #'
      );
    });

    test('formatURLConditional is idempotent', function (assert) {
      const url = formatURLConditional('https://emberjs.com/logos#Branding');

      assert.strictEqual(
        formatURLConditional(url),
        'https://emberjs.com/logos/#Branding',
        'We get the same URL back.'
      );
    });
  });

  module('When ? is in the URL', function () {
    ...
  });

  module('When neither # nor ? is in the URL', function () {
    ...
  });
});
```

I don't think we need a nested module for `formatURLConditional` (as shown in line 5), since the utility currently exports only 1 function.

I suggest that we also test that `formatURLConditional` is idempotent. That is, when you run the function again, we get the same result:

```
f(f(x)) === f(x)
```

ijlee2 (commented):
I tested on the Netlify deploy that the filtering of mascots works, and that the `#` (fragment) in survey pages works also.

Before merging the pull request, I'd like to clean up utility and test for maintenance. Can you have a look at my suggestions?

fozy81:
@ijlee2 I've added your suggestions, really appreciate your feedback and explanation. Let me know if I've missed anything.

ijlee2 (approved):
Looks great, thank you for working on this!

PR #878, opened 2021-11-24
Remove release timeline

wagenet:
It's always out of date and frankly it just looks embarrassing.

I should note that I still like the idea of the release timeline but this seems better than the current state. I also admit to this being a somewhat quick and rough change.


MelSumner (approved):
I have no objections to this; if someone else does I suppose they can just add it back later.

kategengler:
Please merge this! It is _never_ right and we've had a number of attempts to rework this page but they've never gone through. There is never exactly 6 betas (there's at least a beta a week iff there is anything to backport).

wagenet:
@MelSumner @kategengler I don't have access to approve in Percy, but I can force-merge since I'm an admin on this repo.

MelSumner:
I'll work on it if we notice any weird stuff. 👍

PR #884, opened 2021-12-17
Update browser-support

jaredgalanis:


jaredgalanis:
not stale

chriskrycho:
@jaredgalanis @locks this is a great and much-needed update—can we get it finished up and pushed across the line? 🤩

chriskrycho:
@locks and I talked about this via DM a bit and we don’t have a *perfectly* clear idea what the right move here should be. I *suspect* that instead of listing specific versions for the evergreen browsers, we should just say the policy there, and list specific versions for the browsers which *do* have specific versions (read: Safari).

mansona (approved):
Looks good to me 👍

PR #889, opened 2022-01-26
updated typo in paragraph

whoami-shubham:
replaced `too—` with `to` in this paragraph
<br/>
image

### DIFF
--- a/app/templates/index.hbs
+++ b/app/templates/index.hbs
@@ -43,7 +43,7 @@
       </div>
       <div class="lg:col-2">
         <h3 class="text-lg">Build pipeline</h3>
-        <p>Ember CLI is the backbone of modern Ember apps, providing code generators to create new entities and putting the necessary files in the right place, every time. Ember apps come with a <strong>built-in development environment</strong> with fast rebuilds, auto-reload, and a test runner! <br><br> Ready to deploy? That's easy, too— build your app for production with a single command.</p>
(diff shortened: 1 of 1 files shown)


app/templates/index.hbs
```diff
@@ -43,7 +43,7 @@
       </div>
       <div class="lg:col-2">
         <h3 class="text-lg">Build pipeline</h3>
-        <p>Ember CLI is the backbone of modern Ember apps, providing code generators to create new entities and putting the necessary files in the right place, every time. Ember apps come with a <strong>built-in development environment</strong> with fast rebuilds, auto-reload, and a test runner! <br><br> Ready to deploy? That's easy, too— build your app for production with a single command.</p>
+        <p>Ember CLI is the backbone of modern Ember apps, providing code generators to create new entities and putting the necessary files in the right place, every time. Ember apps come with a <strong>built-in development environment</strong> with fast rebuilds, auto-reload, and a test runner! <br><br> Ready to deploy? That's easy, to build your app for production with a single command.</p>
```
locks, line 46:
```suggestion
        <p>Ember CLI is the backbone of modern Ember apps, providing code generators to create new entities and putting the necessary files in the right place, every time. Ember apps come with a <strong>built-in development environment</strong> with fast rebuilds, auto-reload, and a test runner! <br><br> Ready to deploy? That's easy, too! Build your app for production with a single command.</p>
```

locks (changes requested):
I left a suggestion of what I think we should go with, what do you think?
The original is correct, but I agree that the emdash is very strange, especially with how the text breaks after the emdash.

app/templates/index.hbs
whoami-shubham: yes, that's why I thought ` too—` might be a typo. 
after replacing   `too—` with  `too!` it looks fine.

locks (approved):

locks:
Appreciate the help :)

PR #891, opened 2022-01-26
Fix team name

locks:
Slight adjustment to match our current verbiage around core teams.


rwwagner90 (approved):

MelSumner (approved):
:shipit:

mansona:
I restarted tests because it looks like Percy was down temporarily

PR #896, opened 2022-02-14
Adding IE11 to compilation targets

locks:
At the moment the website's navigation does not work with JavaScript disabled.
This is conjunction with the lack of support for IE11 feels to me too punitive of people on older browsers and a matter of accessibility.

While this change won't help in scenarios where JavaScript is disabled, it should enable older browser to access our main website. The Guides still have IE11 in their targets, as does API Docs.

### DIFF
--- a/config/targets.js
+++ b/config/targets.js
@@ -14,12 +14,12 @@ const browsers = [
 // If you need IE11 support on a version of Ember that still offers support
 // for it, uncomment the code block below.
 //
-// const isCI = Boolean(process.env.CI);
-// const isProduction = process.env.EMBER_ENV === 'production';
-//
-// if (isCI || isProduction) {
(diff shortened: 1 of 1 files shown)


acorncom:
@locks I like the sentiment but for an additional 250k+ of JS and a lot of future headaches with Ember 4 I’m not sure it’s worth it. Seems like we could argue that no devs are going to be using IE 11 on their machines and move on? When Microsoft is busy killing it I don’t think we want to be supporting it …

MelSumner:
JS is required for accessibility & we don't support IE. 🤷‍♀️

MelSumner (changes requested):
Do we have any reports from users who are accessing the site from IE11? Ember no longer supports IE11 so it does not make sense for the website to do so either.

PR #913, opened 2022-03-04
Formally alumnize Igor

locks:
Thank you for everything ❤️


MelSumner (approved):
Since Igor let us know that he was going to take a break from all tech in general, I'll go ahead and approve this as an exception (usually we prefer the alumnized person set up the PR or approve it themselves).

PR #925, opened 2022-03-17
Update survey and banner callouts

jeffdaley:
CleanShot 2022-03-17 at 16 07 13@2x

- Moved the Community Survey callout from the top bar to the homepage hero section
- Merged the BLM and Ukraine components into a simple `<CalloutBanner>` component.


app/components/callout-banner.hbs
```diff
@@ -0,0 +1,17 @@
+{{#if this.isShown}}
+  <div class="callout-banner">
```
ijlee2, line 2: Can we a test selector to help write tests?

```suggestion
  <div
    class="callout-banner"
    data-test-callout-banner
  >
```

tests/integration/components/callout-banner-test.js
```diff
@@ -0,0 +1,32 @@
...
+    await render(hbs`
+      <CalloutBanner>
+        I have an announcement to make
+      </CalloutBanner>
+    `);
+
+    assert
+      .dom('.callout-banner')
+      .exists()
+      .hasText('I have an announcement to make');
```
ijlee2, line 19: `hasText()` is a stronger assertion than `.exists()` is. This means, we can remove the latter (a weaker assertion) and still have the same test coverage.

(Aside: Here, the terms "strong" and "weak" come from math when writing a proof.)

```suggestion
      .dom('[data-test-callout-banner]')
      .hasText('I have an announcement to make');
```

I'd also recommend using a `data-test-*` instead.

Related: https://github.com/ember-learn/ember-website/pull/925/files#r832893092

tests/integration/components/callout-banner-test.js
```diff
@@ -0,0 +1,32 @@
...
+
+  test('callouts can be made dismissible', async function (assert) {
+    await render(hbs`
+      <CalloutBanner @dismissible={{true}}>
+        You can dismiss this announcement
+      </CalloutBanner>
+    `);
+    assert.dom('.callout-banner').exists();
+    await click('.toggle-close');
+    assert.dom('.callout-banner').doesNotExist();
```
ijlee2, line 30: Let's prefer a `data-test-*` over a CSS class. I also recommend providing an easy-to-read description for each assertion, to help other contributors understand what scenario the assertions cover.

```suggestion
    assert
      .dom('[data-test-callout-banner]')
      .exists('The user sees the callout banner.');

    await click('[data-test-button="Close"]');

    assert
      .dom('[data-test-callout-banner]')
      .doesNotExist('The user should not see the callout banner.');
```

ijlee2 (commented):
Thanks for abstracting a banner component and creating the `<Index::HeroCallout>`. The design for the latter looks great.

I have a few suggestions regarding writing tests. If you can include the changes, I think that will help with future maintenance.

MelSumner (approved):

PR #931, opened 2022-04-08
Update header-links.js

MelSumner:
Removes link from navbar since it was placed back in the banner.


jenweber:
Thankd for this PR! It seems there is one test that asserts the number of links present, and it should be reduced by 1:
Screen Shot 2022-04-11 at 12 15 35 AM

Once that test is updated, this is good to go.

PR #934, opened 2022-04-21
feat(ecs-data): add data + convert hbs to publication

josephdsumner:
If merged, this pull request will publish the initial results page for the 2022 Ember Community Survey. 

The primary changes are:
- the controller containing all of the data + charts has been added
- the template currently hosting the survey announcement has been updated to publish the results

There are a few known TODOs documented for dataset cleanup as well as the addition of the text content summarizing / interpreting the results. 

Update -- TODOs per reviews:

Charts to Update: @josephdsumner 
- [x] Add 'How long have you been using Ember?' question
- [x] Update 'Other' data: preferred IDE
- [x] Update 'Other' data: work sectors
- [x] Update 'Other' data: test frameworks
- [x] ~~Update 'Other' data: server-side frameworks~~ cannot do this one, as data missing
- [x] Update 'Other' data: deployment

Other changes:
- [ ] Appropriately accommodate that we had two sponsors this year
- [ ] Color palates (there are some already in the controller that were commented out to keep the linter happy)
- [ ] Content + remarks interpreting the results, where appropriate @MelSumner 
- [ ] Remove the Series Label from single-series charts (e.g. '2022')

PR #935, opened 2022-04-22
Refactored chart utilities, conditionally hid the chart legend

ijlee2:
## Description

A feature was requested in #934. Namely, when a chart doesn't have multiple series, we hide the legend because the context is assumed to be clear.

I updated the utility files so that the feature will apply to all survey charts from 2015 to 2022.

## How to Review

The pull request can be divided into 3 parts. It may help to toggle the `Hide whitespace` option on GitHub.

- Commits 1-5: Used the `beforeEach` hook to use a common set of data across tests. Updated test descriptions and comments so that they may make more sense to a new contributor.
- Commits 6-10: Declared `chart` and `rawData` as tracked properties so that Ember can recompute values that depend on these properties.
- Commits 11-14: Allowed charts to hide legend when `series` has 0 or 1 element(s). (The `legend` key does not apply to pie charts in Highcharts.)

## Screenshots

Before:

The vertical bar chart shows a legend

After:

The vertical bar chart doesn't show a legend anymore

### DIFF
--- a/app/utils/highcharts/area-spline-chart.js
+++ b/app/utils/highcharts/area-spline-chart.js
@@ -13,12 +13,16 @@ export default class AreaSplineChart {
   }
 
   get highchartsOptions() {
-    const { chart, series } = this;
+    const { chart, isLegendEnabled, series } = this;
 
     return {
       chart: {
         backgroundColor: 'transparent',
         type: 'areaspline',
+      },
+
+      legend: {
+        enabled: isLegendEnabled,
       },
 
       series,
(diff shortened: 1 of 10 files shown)


MelSumner (approved):
This looks great. Thank you for doing this work. 👍

josephdsumner (approved):

josephdsumner:
Looks great, thanks for adding this feature! It is great to have this flexibility for the recently-added questions where no historical series is available.

josephdsumner:
Took a sneak peak at this feature rebased onto the new 2022 results page and it looks good to me!

ecs-2022-label

PR #936, opened 2022-04-25
Alumnize Chris Garrett (and update name)

pzuraq:


chriskrycho (approved):
Thanks for the incredible work you did in the time you served on the Framework team! 💙

PR #939, opened 2022-05-19
Refactored <Highcharts> component

ijlee2:
## Description

Prior to this pull request, we made use of `@ember/render-modifiers` to render svg elements. Instead, we can use `ember-modifiers` and make `<Highcharts>` a template-only component.

## How to Review

I recommend reviewing the commits one by one. Each commit makes a small, focused change.

- Commits 1-2, 9: Managed dependencies
- Commits 3-6: Created the `{{draw-chart}}` modifier. The modifier acts as a bridge between Ember and Highcharts.
- Commit 7: Updated the `<Highcharts>` component. Thanks to the modifier approach, we do not need a backing class.
- Commit 8: Fixed failing tests. Added application tests for the 2022 Ember community survey page.


app/modifiers/draw-chart.js
```diff
@@ -0,0 +1,69 @@
...
+
+    registerDestructor(this, this.destroyChart.bind(this));
+  }
+
+  initializeHighcharts() {
+    highchartsAccessibilty(Highcharts);
+
+    this.highcharts = Highcharts;
+    this.highcharts.setOptions(optionsForAllCharts);
+  }
```
ijlee2, line 45: Previously, we had dynamically imported Highcharts, then observed that the import is finished before rendering an svg element.

Although it's possible to (in a modifier) dynamically import Highcharts and its extra packages (like the accessibility module), I found that it's easier for others to debug and maintain code when we do static import.

app/modifiers/draw-chart.js
```diff
@@ -0,0 +1,69 @@
...
+      {
+        chart: {
+          events: {
+            render: () => {
+              element.setAttribute('data-render-state', 'settled');
+            },
+          },
+        },
+      },
+      chart.highchartsOptions
```
ijlee2, line 60: In a future pull request, we'll rename `highchartsOptions` to `options`.

tests/acceptance/survey/2016-test.js
```diff
@@ -28,7 +28,10 @@ module('Acceptance | survey/2016', function (hooks) {
   test('Accessibility audit', async function (assert) {
     await visit('/survey/2016');
     await waitUntilAllChartsAreDrawn();
-    await a11yAudit();
+    await a11yAudit({
+      exclude: [['[data-test-chart]']],
+      include: [['#ember-testing-container']],
+    });
```
ijlee2, line 34: We'll assume that Highcharts does create an accessible chart. The svg element, which contains many elements, can slow down an application test (by an order of 0.1s - 10s).

tests/integration/components/highcharts-test.js
```diff
@@ -1,46 +1,20 @@
...
-        {
-          color: '#F1BF28',
-          label: 'JavaScript',
-          values: [2.7, 35.9, 61.4],
-        },
-      ],
-    });
+    this.chart = {
+      highchartsOptions: {},
+    };
```
ijlee2, line 14: I simplified the test setup. We're not interested that a particular chart works (better tested through unit tests for the `VerticalBarChart` class), but that the component works in general.

ijlee2:
@jenweber This pull request is ready for review. Can you check that:

- The deploy website continues to render the Ember community survey pages.
- Percy snapshots are approved. (The before screenshots are out-of-sync. I think Percy didn't run in a recent pull request because we were out of credits.)

app/modifiers/draw-chart.js
ijlee2: The accessibility module allows users to use their keyboard to interact with charts. We can also pass chart captions and ARIA labels to better describe axes.

jenweber (approved):
Thank you so much for this work! It was easy to follow along with thanks to your commits + comments. I click tested the latest survey year on the test app, and searched the snapshots for any difference that was more than the page shifting down. I did not see any visual changes, which is good. Including the accessibility module is very nice too.

This would be a nice foundation for a conference talk someday. I have never really understood how to use modifiers directly, without render-modifiers, and I learned a lot just now. Bookmarking this PR for later reference.

ijlee2:
@jenweber Appreciate your feedback! I'm glad that the sequence of commits helped with understanding the pull request.

PR #945, opened 2022-07-15
Fix typo for modifying

chrisrng:
Fix modifing to modifying


MinThaMie (approved):
Thanks for catching :)

PR #948, opened 2022-07-28
Adding Wednesday Solutions to users

aliHafizji:
Adding Wednesday (wednesday.is) to the list of users. We're an agency actively using ember on client projects.


jenweber (approved):
Thank you for the PR!

PR #961, opened 2022-09-16
simplabs is rebranding to Mainmatter

marcoow:
We're rebranding the company – still same team, same work etc. but under a new name and new visual brand 🎉 

This updates all mentions of simplabs to Mainmatter.

We're releasing the rebranding on Sunday 18th so would be great if this could be merged then or in the morning of the 19th.


mansona (approved):
Looks good 🎉

PR #989, opened 2023-02-06
Fix TypeScript case

axsuul:


MinThaMie (approved):
Thanks for catching this :)

PR #991, opened 2023-02-08
Remove idobata from users

fossamagna:
This service is closed.


MinThaMie (approved):
The mentioned URL does not seem to work indeed :) Thanks for noticing!

PR #996, opened 2023-03-02
Remove Nedap as user

MinThaMie:
Since we no longer use Ember at Nedap (and haven't really used it much) I think it would be better if we removed it.


MelSumner (approved):
:shipit:

PR #1006, opened 2023-03-27
update ember-asset-size-tracker

mansona:
This brings it up to the latest and at also allows it to always post a comment even when the creator of the PR does not have merge permissions 👍


MinThaMie (changes requested):
Since it's erroring in CI I'll not approve this :)

mansona:
@MinThaMie that's fixed now btw 👍

MinThaMie (approved):

Issue #1007, opened 2023-03-29
Copy/Content Updates to /logos/ page


wifelette:
Hello!

Simple issue for anyone who has time to implement before I get to it. These are some copy updates that were written down a long while ago but never implemented, and I found them when tidying the Notion 😛 I'm here for any questions!

### First Tasks

- [ ]  Make each headline on the Logos page an anchor link, since we frequently need to link directly to them, and so the new copy below can reference them.
- [ ]  Make an anchor link to each entry on the /mascots page, so we can link to them from the copy below.
- [ ] Add the copy below
- [ ] Add links to each instance of _Standard Art Set_
- [ ] Add the correct links to each item listed in the _Standard Art Set_ section

### New Copy/Sections to Add

This new content belongs under the Branding Page section titled _The Ember Mascots_, between it and the _Naming_ section:

> ### Company Technology Pages
> 
> Many companies choose to display icons and logos of their technology stack on their company About Us or Career pages. The **Tomster Lockup, Classic Zoey** and **Primary Logo** are approved for this use case. Specific permission is not required. The images should link to [https://emberjs.com/](https://emberjs.com/).
> 
> ### Blog Posts and Articles
> 
> Blogs by individuals, companies, and media publications may utilize any of the pieces of art below in the _Standard Art Set_ section, provided they add the following footer content to their posts/pages:
> 
> > *Ember, the Ember logo, and the Tomster/Zoey designs are exclusive trademarks registered in the United States by Tilde Inc, and are used here with permission. [Blog page/product/etc] is unaffiliated with the Ember project, and the use of this art does not imply any endorsement or official status.*
> 
> We reserve the right to require the art to be taken down at any point, with or without explanation.
>
> ### Presentation Slides
> 
> The _Standard Art Set_ may also be used in presentation slides for industry conferences, internal corporate presentations, training presentations, and funding pitch decks, provided the following footer/asterisk content is added to a slide earlier in the Deck than where the art is used (it's alright for it to be small). It needs to appear once per deck, not every time a piece of art is used:
> 
> > *Ember, the Ember logo, and the Tomster/Zoey designs are exclusive trademarks registered in the United States by Tilde Inc, and are used here with permission. [Blog page/product/etc] is unaffiliated with the Ember project, and the use of this art does not imply any endorsement or official status.*
>
> ### Standard Art Set
> The following items are included in the _Standard Art Set_:
> 
[...]

PR #1009, opened 2023-04-14
Add /logos/ route new sections and anchor links

samridhivig:
This MR works on the Copy/Content Updates to /logos/ page #1007 issue tasks:

1. Each section has an ID that can be used in the anchor link in order to navigate to a specific section
2. Added id to each Mascot ListItem li i.e `mascot.id`
3. Added the necessary sections
4. Added links to each instance of the Standard Art Set
5. Added the links to the mascot items specified in Standard Art Set

### DIFF
--- a/app/components/mascots/mascot-list.hbs
+++ b/app/components/mascots/mascot-list.hbs
@@ -1,6 +1,6 @@
 <ul class="unstyled grid {{if (eq @display "small") "sm:grid-2 lg:grid-3" "sm:grid-1 lg:grid-2"}}">
   {{#each @mascots as |mascot|}}
-    <li data-test-mascot>
+    <li id={{mascot.id}} data-test-mascot>
       <Mascots::MascotList::Item
         @mascot={{mascot}}
       />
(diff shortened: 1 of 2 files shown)


mansona (changes requested):
Thanks for the contribution 🎉 can you just remove the yarn.lock and we can get this merged? we're just using npm on this project so we don't have a yarn.lock file here 👍

wifelette:
It looks like the anchors on the Mascots aren't working. If you follow the new links on the Brand page, the URL is updating, but the mascots page jumps back to the top. Just me?

This looks otherwise amazing, thank you @samridhivig so much for doing the work, and sorry I've been offline and haven't said that sooner!

MinThaMie:
Same here, in the preview and running it locally ;) So theoretically this should work but the router inserts a `/` between the page and the section so that's why it jumps to the beginning. I'll check at the learning meeting if somebody there knows why.

mansona:
So this came up a long time ago when I tried to implement the same thing. The issue is that we need to enable an experimental feature on Fastboot that will allow this to work as expected 🙈 (It's EXPERIMENTAL_RENDER_MODE_SERIALIZE for reference)

We discussed this in the Learning team today and we think it's probably best to update this PR to not include the links to the specific Tomsters/Zoes and we can fix the linking problem in a different PR. Essentially we're not 100% comfortable enabling the experimental feature without testing it and we didn't want to block the other important bits of this PR.

What do you think?

mansona:
It turns out we figured out how to make the links work with #1120 so I've rebased this and if the links are working on the demo site I'll get it merged 👍

mansona (approved):

PR #1034, opened 2023-06-08
Embroider Initiative Sponsors

mansona:
This needs to be a) finished and b) validated before merging


wifelette (changes requested):
Looks great! It's missing a period after "Read more about the Embroider Initiative and see all the contributors [here](https://mainmatter.com/embroider-initiative/)" but otherwise, ship it!

PR #1041, opened 2023-07-04
New Link: "Try" (out the framework)

NullVoxPopuli:
We can have people try out ember in their browsers before they do any setup on their machines.

Note, right now booting up broccoli in the browser is not the fastest experience.
But the work on embroider, getting to vite, and converting the core addons to the v2 addon format will cut that time down _immensely_

### DIFF
--- a/app/utils/header-links.js
+++ b/app/utils/header-links.js
@@ -1,5 +1,10 @@
 export default [
   {
+    name: 'Try',
+    type: 'link',
+    href: 'http://new.emberjs.com',
+  },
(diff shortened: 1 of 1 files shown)


locks:
@NullVoxPopuli we _just_ synced the navbar across the variouos apps xD Can you submit this to ember-styleguide instead?

nickschot:
For your convenience: https://github.com/ember-learn/ember-styleguide/blob/master/addon/constants/links.js

Issue #1051, opened 2023-08-27
[BUG] Brand Guide URL 404s


wifelette:
**Describe the bug**
On `/logos` aka `/branding`, there is a link to the PDF brand guide. It goes to `https://emberjs.com/pdfs/Ember-Brand-Guidelines.pdf/` instead of `https://emberjs.com/pdfs/Ember-Brand-Guidelines.pdf`. The trailing slash produces a 404 instead of the Guide, which is otherwise there and working.

<img width="477" alt="image" src="https://github.com/ember-learn/ember-website/assets/25433/456972a4-ff1f-42c2-a10b-35b0f4f57488">

**To Reproduce**
1. Go to `https://emberjs.com/logos/`
2. Click on Download Brand Guidelines

anshikjain18:
Hello @wifelette, I would like to work on this issue. From what I can see opening the 'Download brand guidelines (PDF)' link in a new tab works fine but if we try to open it in the same tab it gives 404. We can set the anchor tag to always open the link in a new tab so that the branding page is still present when we click the link. This solution is recommended but if the requirements are to open the link in the same tab that can also be done.

wifelette:
That sounds great to me! So long as we aren't 404ing anymore :) That said, happy for anyone else on the Learning Team with more current/specific technical thoughts to jump in. Whatever gets the job done, and thanks for stepping up to help!

rameshvoodi:
hello @wifelette I have made a PR ##1052 that can solve this issue, please look into it.

PR #1052, opened 2023-08-27
fix: Fixed 404 ERROR

rameshvoodi:
##What does this PR do?

Fixes #1051 
Closes https://github.com/ember-learn/ember-website/pull/1054
Closes https://github.com/ember-learn/ember-website/pull/1053

##Changes made

at *ember-website\app\templates\logos.hbs* added *target="_blank* to <a> tag and changes *href* to **/pdfs/Ember-Brand-Guidelines.pdf**

##video

https://github.com/ember-learn/ember-website/assets/53528683/a6e2e8c2-6aae-4a8b-a076-8e24984b0b37

PR #1053, opened 2023-08-27
[BUGFIXED] Brand Guide URL 404s

pipoarks:
Hello @wifelette,
 #1051

I have made a PR  that can solve this issue, please look into it and Advice.

Thanks in advance.

### DIFF
--- a/app/templates/logos.hbs
+++ b/app/templates/logos.hbs
@@ -3,7 +3,7 @@
   <section aria-labelledby="branding" class="branding-section">
     <h1 id="branding">Branding</h1>
     <p>The Ember brand and associated assets are part of a meticulously curated experience. We take great pride in what our team and community have helped bring to life, and ask that you follow our guidelines to ensure consistency across all Ember properties.</p>
-    <p><a href="/pdfs/Ember-Brand-Guidelines.pdf">Download brand guidelines (PDF)</a></p>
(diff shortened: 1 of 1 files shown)


app/templates/logos.hbs
```diff
@@ -3,7 +3,7 @@
   <section aria-labelledby="branding" class="branding-section">
     <h1 id="branding">Branding</h1>
     <p>The Ember brand and associated assets are part of a meticulously curated experience. We take great pride in what our team and community have helped bring to life, and ask that you follow our guidelines to ensure consistency across all Ember properties.</p>
-    <p><a href="/pdfs/Ember-Brand-Guidelines.pdf">Download brand guidelines (PDF)</a></p>
+    <p><a href="https://emberjs.com/pdfs/Ember-Brand-Guidelines.pdf">Download brand guidelines (PDF)</a></p>
```
NullVoxPopuli, line 6: isn't https://github.com/ember-learn/ember-website/pull/1052 the fix?

if we add a domain here, then we can't see preview versions of this pdf.
I think we want to be able to view subdomain versions of the pdf, like in the other PR?
image

NullVoxPopuli (commented):
Thanks for submitting, but I think https://github.com/ember-learn/ember-website/pull/1052 is what we want <3

PR #1059, opened 2023-09-11
Update browser-support.hbs to include v5

SergeAstapov:
https://emberjs.com/browser-support/ states v3 in present tense and talks about v4 in future tense, which is outdated.

Also v5 has been released some time ago but page does not mention it.

### DIFF
--- a/app/templates/browser-support.hbs
+++ b/app/templates/browser-support.hbs
@@ -3,6 +3,39 @@
   <section aria-labelledby="section-browser-support-policy">
     <h1 id="section-browser-support-policy">Ember.js Browser Support Policy</h1>
 
+    <h2>Ember 5.0.0</h2>
+
+    <p>
+      In Ember 5.0.0, the framework supports the following major browsers:
+    </p>
+
+    <ul class="list-unstyled layout my-3">
+      <EsCard class="lg:col-2">
+        <div class="text-center text-md">Desktop</div>
+        <ul>
+          <li>Google Chrome >= 103</li>
+          <li>Mozilla Firefox >= 102</li>
+          <li>Microsoft Edge >= 110</li>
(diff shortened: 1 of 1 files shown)


app/templates/browser-support.hbs
```diff
@@ -3,32 +3,10 @@
...
-            <li>Microsoft Edge</li>
-            <li>Internet Explorer 11</li>
-            <li>Safari</li>
-          </ul>
-        </div>
-      </div>
-    </div>
-
-    <h2>Ember 4.0.0</h2>
+    <h2>Ember 4.0.0 and Ember 5.0.0</h2>
```
SergeAstapov, line 6: moved v4 and v5 to the top as that's what I think most ppl would like to see first

MinThaMie (approved):
Awesome :)

kategengler:
This seems fine to me but might miss some of the nuance of the browser support policy that started with v4. I am surprised to find #884 never landed.

SergeAstapov:
@kategengler totally agree! I was pretty sure I would miss some details, but seems like page in current state makes bad impression of being neglected and having some details missed is better than not having them at all.

My assumption would be: if there is no wrong content in here and structure looks good, land it and iterate to clarify and refine the details.

kategengler:
I think that's fine but do still wonder why #884 was never landed. 

For future reference, browser support is a rolling window for the evergreen browsers that can be updated in each minor. Current support can be seen here https://github.com/emberjs/ember.js/blob/main/lib/browsers.js and this is also exported on Ember's index, so in node `require('ember').supportedBrowsers` gets the list. 

This is the relevant RFC https://rfcs.emberjs.com/id/0685-new-browser-support-policy/

mansona (approved):
Looks good to me 🎉

PR #1073, opened 2023-12-18
Update LTS page for data breaking lockstep

jaredgalanis:


mansona (changes requested):
Sorry I didn't give my feeback via text last time 🙈 I'll write them down this time so we don't forget 😂 

- there should be one table each per project and have headers that tell you what one they are for (ember-source or ember-data)
- the heading for each table should be LTS version, Bugfixes until, Security patches until, and Status where status says `end of life` or `active`

mansona (approved):
Looks great! LGTM 💪

PR #1076, opened 2024-01-08
Update to Ember 4.12.2

simonihmig:
This updates the website to Ember 4.12.2, including:
* the app boilerplate and related dependencies by running ember-cli-update
* updated and deduped transitive deps (due to issues) by regenerating the lockfile (except for axe-core/a11y-testing, which seems to have breaking changes in a minor version update)
* ember-data to 4.12, including FastBoot fixes (https://github.com/emberjs/data/issues/8475)


app/routes/community/meetups/index.js
```diff
@@ -7,4 +7,11 @@ export default class CommunityMeetupsIndexRoute extends Route {
   model() {
     return this.store.findAll('meetup');
   }
+
+  afterModel() {
+    // ember-data runs flushAllPendingFetches using setTimeout, so outside of the Ember runloop.
+    // Under prember this causes `Attempted to call store.adapterFor(), but the store instance has already been destroyed.`
+    // So wait a bit
+    return new Promise((resolve) => setTimeout(resolve, 10));
+  }
```
simonihmig, line 16: This is an ugly workaround for an issue with ember-data and FastBoot/prember. e-d calls `flushAllPendingFetches` with setTimeout (which Ember's settledness does not capture, I believe that's the problem), but as the app is torn down immediately after rendering in prember, we are getting: 

```
Assertion Failed: Attempted to call store.adapterFor(), but the store instance has already been destroyed.
```

I was not really willing to spend much time investigating a "proper" solution, as

1. the issue might have been resolved in new releases maybe
2. I think we should actually get rid of e-d here. I'll raise a separate issue to explain my thoughts...

IgnaceMaes: One concern I have is that with the current context in the code comment is that it makes it quite daunting to ever touch this part of the code. Is it possible to link to a bug ticket. (Is this even considered a bug, it seems intentional?) Or state when/how this workaround could be removed.

Removing ED would solve the problem as well, but is out of scope for this PR.

Another alternative could be downgrading Ember Data to 4.11 I assume, if this "issue" got introduced in 4.12?

IgnaceMaes:
Thanks for this!

I wanted to look into upgrading as well and only then noticed you had made a PR already 🎉

app/routes/community/meetups/index.js
simonihmig: I didn't file an issue, because I wouldn't want to do so for a version that is out of date and without knowing that the problem is still present in the most recent one. But also ED has deprecated FastBoot support, so likely that could mean they wouldn't even consider it a bug?

Tbh, I don't really remember what happened when downgrading to ED to 4.11, but I am pretty sure I tried. This work is already two months old, which hits my brain's capacity limit 😅 

The path I would be pursuing is to [remove ED here](#1078 ), so the code you are concerned with wouldn't be around for too long!

mansona: I'm also a bit concerned about this hack being merged 🤔 I accept your intention to potentially remove ember-data since it's such a simple request response mode, but the intention was for the website to be a good **simple** demonstration of the capabilities in the default Ember blueprint. I think we should dig a bit deeper into this issue and potentially draft some help from the Ember data team especially if this issue is potentially affecting other people relying on ember-data and prember

Windvis: I created https://github.com/ember-learn/ember-website/pull/1193 and it seems the issue is no longer there now. Not sure what actually fixed it, but I was able to build the app without issues and without code changes.

PR #1084, opened 2024-02-12
Correct number of weeks added to promotion date

kategengler:
It would be 36 added to the release date, but only 30 from promotion date


MinThaMie (approved):
Trust you on this :)

PR #1088, opened 2024-02-23
Update the browser support policy page

NullVoxPopuli:
RFC: https://github.com/emberjs/rfcs/pull/984

### DIFF
--- a/app/templates/browser-support.hbs
+++ b/app/templates/browser-support.hbs
@@ -3,6 +3,40 @@
   <section aria-labelledby="section-browser-support-policy">
     <h1 id="section-browser-support-policy">Ember.js Browser Support Policy</h1>
 
+    <h2>Ember 6.0.0</h2>
+
+    <p>
+      In Ember 6.0.0, the framework supports the following major browsers:
+    </p>
+
+    <ul class="list-unstyled layout my-3">
+      <EsCard class="lg:col-2">
+        <div class="text-center text-md">Desktop</div>
+        <ul>
+          <li>Google Chrome >= 103</li>
+          <li>Mozilla Firefox >= 102</li>
+          <li>Microsoft Edge >= 110</li>
+          <li>Safari >= 16.4</li>
+        </ul>
+      </EsCard>
+      <EsCard class="lg:col-2">
+        <div class="text-center text-md">Mobile</div>
+        <ul>
+          <li>Google Chrome >= 112</li>
+          <li>Mozilla Firefox >= 110</li>
+          <li>Safari >= 16.4</li>
+        </ul>
+      </EsCard>
+      <EsCard class="lg:col-2">
+        <div class="text-center text-md">Testing</div>
+        <ul>
+          <li>Google Chrome</li>
+          <li>Mozilla Firefox</li>
+        </ul>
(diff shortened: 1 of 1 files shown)


app/templates/browser-support.hbs
```diff
@@ -3,6 +3,40 @@
   <section aria-labelledby="section-browser-support-policy">
     <h1 id="section-browser-support-policy">Ember.js Browser Support Policy</h1>
 
+    <h2>Ember 6.0.0</h2>
+
+    <p>
+      In Ember 6.0.0, the framework supports the following major browsers:
```
NullVoxPopuli, line 9: ember 6 isn't out yet, so I'm not sure what we want to do here -- or maybe this whole section could be omitted until ember 6?

especially since the versions below on lines 16-19 are likely to change

app/templates/browser-support.hbs
```diff
@@ -109,13 +143,15 @@
           <li>Google Chrome</li>
           <li>Mozilla Firefox</li>
           <li>Microsoft Edge</li>
+          <li>Safari</li>
```
mansona, line 146: Based on my undersanding of the RFC, the change for safari will only affect ember@6 so we should not be changing any other section on this page other than adding the ember@6 section 👍

mansona (dismissed):
one comment but I'm also "requesting changes" because we won't be merging this until ember@6 is released. We can keep this PR to remember that we need to merge it but it should be added to some sort of list of things to do for v6 👍

app/templates/browser-support.hbs
NullVoxPopuli: yup! apologies, I went too fast here

NullVoxPopuli: but this is the evergreen section -- so since we're waiting to merge until ember v6, wouldn't it be true that safari is evergreen?

mansona: so this is the evergreen section for ember@3 back when there was a bigger distinction between the two cases. For each of the newer Ember versions above this we include Safari 12 on both mobile and desktop even though it's not evergreen. 

We could legitimately restructure this page generally to explain the policy first and make sure that each specific version has an evergreen and non-evergreen list 🤔

NullVoxPopuli: oh! that makes sense. and the h3 vs h2 above (mentioning ember v3) explains that as well!

achambers:
https://github.com/NullVoxPopuli/ember-website/pull/2

mansona (changes requested):
@NullVoxPopuli mentions that he's not sure what to do about the future reference to an unreleased Ember version. I think we need to be very careful about how we phrase this. I'm requesting changes on this but I'm also committing to pair with @NullVoxPopuli to get the phrasing on this right 👍

app/templates/browser-support.hbs
kategengler: I think maybe omit this section until 6.0 exists -- the policy below is enough

mansona (approved):
This is significantly improving the whole page and now effectively links to the future changes in the support policy 🎉

jaredgalanis (approved):

PR #1090, opened 2024-03-08
chore: delete meetups that refer to discontinued groups

jahrock:
Unfortunately, there are several discontinued meetup groups.

Of course, it would be nice if we could reactivate some of these groups. However, so that the active groups don't get lost in the crowd, I have removed the non-existent groups for the time being.


NullVoxPopuli (approved):
Will be good to not mislead people as they browse this page.

Also, I'm not sure meetup is used much these days. Feels like I haven't had meetups marketed to me much since covid :shrug:

IgnaceMaes (approved):
Looks good, thank you for the contribution! 🙌 

EDIT: looks like there's a tiny lint issue. Could you run `lint:js:fix`?

PR #1093, opened 2024-03-11
fix: typo in pix user

VincentHardouin:


IgnaceMaes:
That's better 😬 

Thanks!

PR #1108, opened 2024-05-04
combo of npm + ignore-scripts=true means the app can't start -- solution, which is a no-op for folks who don't change this setting, is to undisable scripts

NullVoxPopuli:
via the local `.npmrc` file


mansona:
this is fixed now 👍

PR #1137, opened 2025-03-25
Update ota-insight user to lighthouse

ArnaudWeyts:
We rebranded a while ago, but the users page was never updated.

Source: https://www.mylighthouse.com/resources/blog/ota-insight-rebrands-as-lighthouse-to-illuminate-new-capabilities-and-launch-of-a-unified-commercial-platform


IgnaceMaes (approved):
Looking good 😄

Issue #1158, opened 2025-07-26
Update and improve "Who uses ember" page


TeriyakiBomb:
There’s a bunch of inconsistency and dead/moved/acquired sites in this list. 

**Why this matters?**
Optics! This may or may not be a frequently visited page but the emberjs site shows a huge number of companies. It’s such a good way to reinforce that the framework is relevant, in use and still in use by some big businesses. Dead links just look bad.

https://emberjs.com/ember-users/

**Missing sites**

All of these sites are either no longer online or have been sold to advertising/gambing placeholders. ~I’m happy to put together a PR to remove all of these.~ **#1160** 

* Benefit cloud
* Group talent
* Blue tube
* Twine
* Koh active
* brainistic
* Plantain
* Porkepic
* Karma
* Nosolo software
* Sunspace
* Llano seco
* Data field
* Wagaroo
* Pertino
* Spot on
* Simple reach
* Pivot share
* Caxanuma
* Caliper
* Scan
* Noesis software
* Radflex
* Chronosave
* centroy
* Pluck
* Locbox
* Swink TV 
* Bondable
* huboard
* The working group
* Faster agile
* Balanced
* Unlockable
* Orchestrate
* Speedpos
* Radium
* Mantra
* Typework
* Zing
* Screenlight
* Export to china
* Pushmotex
* Poetic
* Sparehanger
* whichledlight
* Foogi
* Zesty
* Vestorly
* What’s due
* Go zooga
* News ninja
* Monograph
* Privatt
* Shearwater
* Streamrail
* Society of grownups
* Mammoth
* Feeds
* Campus discounts
* Mapc
* 201 created
* Crowdscriber
* NYC Planning
* crx
* netuitive
* poynt
* Khorus
* Samewave
* shyft
* Verizon digital media
* Tourme
* Paycertify
* camilyo
* Ocular
* Digitalmark
* American specialty health
* Scalefactor
* Bengala
* Starshot
* https://egitimilan.com/

**Changed/sold/merged sites**

Not entirely sure what should be done about these, I suspect in most cases, we do nothing. However some of these businesses have been acquired by some pretty interesting larger businesses, so it’s interesting if they’re still making use of ember. However the effort to go through all of the more interesting ones and finding out sounds kind of unfeasible? Not sure really. Open to ideas.

* Funding gates is now https://www.runviably.com/ 
* Yes mail is now https://www.data-axle.com/ 
* Touchay is now https://www.wesleyclover.com/ 
[...]

PR #1159, opened 2025-07-26
Improve Ember users

TeriyakiBomb:
I've filtered out dead users as per the list in #1158 there was already an inactive flag in `data/users` which also contains some users that have live products. At some point the work to mark them inactive has already been done, so this might be pruning the list a little too much. I welcome some feedback on this.

I've also updated the number of featured users to push some of the larger companies further up the list - again I'd love some feedback on who might be missing, who should not and if there's some reason I'm unaware of for those users to be further down the list.

Something I ran into was an error trying to filter inactive with `store.query` returning a 400. I don't have enough insight to know why that's the case, but I've left a comment in `app/routes/ember-users.js` any insight would be fantastic. I already asked @NullVoxPopuli and they said someone in the learn team might know more.

### DIFF
--- a/app/routes/ember-users.js
+++ b/app/routes/ember-users.js
@@ -4,7 +4,18 @@ import { inject as service } from '@ember/service';
 export default class EmberUsersRoute extends Route {
   @service store;
 
-  model() {
-    return this.store.findAll('user');
+  //This returns a 404 - any insight would be useful
+
+  // model() {
+  //   return this.store.query('user', {
+  //     filter: {
+  //       inactive: false
+  //     },
+  //   });
+  // }
+
+  async model() {
+    const users = await this.store.findAll('user');
+    return users.filterBy('inactive', false);
   }
 }
--- a/data/user/201-created.md
+++ b/data/user/201-created.md
@@ -1,6 +1,7 @@
 ---
 url: 'https://www.201-created.com'
 image: 201-created.png
+inactive: true
 name: 201 Created
 added: 2017-06-16T00:56:33.000Z
 ---
--- a/data/user/american-specialty-health.md
+++ b/data/user/american-specialty-health.md
@@ -1,6 +1,7 @@
 ---
 url: 'https://www.ashcompanies.com/'
 image: american-specialty-health.png
+inactive: true
 name: American Specialty Health
 added: 2019-06-20T17:35:42.670Z
 ---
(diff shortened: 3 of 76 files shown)


MinThaMie (commented):
Thanks for this work :) I was wondering, since you made some of the bigger companies featured, do you/we know if they are still using Ember? Otherwise I might want to feature some smaller but active companies like Phorest

TeriyakiBomb:
Hey @mansona,

First point - I had assumed as much, I was just unsure is there was some other magic going on. Either way, I'll revert it for now given the rest of your request

Filtering vs deletion - I had originally intended to just delete them as you've suggested, but I discovered the inactive thing already existed, but seemed out of use, plus some ambiguity about the pre-existing inactive users and how much crossover there is with the list I compiled. When looking into the data, about 40% of the companies on my list were already marked as inactive. Just to confirm - if it's on the list, do we just delete it and ignore the inactive flag for now?

The featured flag already exists and is in use, the ones I've added aren't _that_ contentious - they're all 9 digit businesses. But given the input from @MinThaMie it's probably safer to revert for now and someone to pick up the work of finding out if they still actively use ember. Although Crowdstrike weren't marked as featured, but aren't they fairly prominent ember users? Happy with whatever you all suggest, you have the context, I just see big company names. :)

mansona:
> When looking into the data, about 40% of the companies on my list were already marked as inactive. Just to confirm - if it's on the list, do we just delete it and ignore the inactive flag for now?

So first of all I didn't know that inactive was originally there. From what I can see there is nothing **reading** this flag on a user so we should remove the whole concept of inactive entirely. As for who to delete, the thing that I said is that if the website is down or pointing at spam (like your original issue mentioned) then the user should 100% just be deleted. That should be the first PR and should not be contentious at all 👍  

If there are any other users that have been marked as "inactive" for any other reason then the decision could be considered marginal and we should discuss that in a separate PR or issue 👍 

> The featured flag already exists and is in use

I also didn't know/remember that the featured flag was there 🙈 🫠  I'm ok having a discussion about adding featured to companies but that should be in a separate PR. There should only ever be a single concept in a single PR so that we can have relevant discussions about one idea at a time.

As for people like Crowdstrike I wouldn't worry too much about them 😉 they are on the sponsors page and that page doesn't add `nofollow` so it has a much bigger impact on SEO for them than the users page 😂

TeriyakiBomb:
With you on all of this, I'll crack on with the deletions and open up a new PR with just that. I'll leave the issue open with the broader conversation and we can pick up the featured companies another time

TeriyakiBomb:
#1160 replaces this

PR #1161, opened 2025-07-28
Add node-support page

NullVoxPopuli:


NullVoxPopuli:
The menu seems wholly controlled from a library, which i'd do an ipgrade pr separately after pring the library (which would only be safe to do after the page exists)

> assume this is added because it was discussed that it was needed during a tooling meeting 

No meeting disco, but the desire for such a page does come up constantly, both by ourselves, and the community.

I'm doing a typescript support page next, because that info isn't even written down anywhere.

MinThaMie:
Cool :) Let's move forward with these pages and then update the menu!

PR #1169, opened 2025-08-14
Add Dentally to the ember-users page

linearza:
Add Dentally to the ember-users page.

https://www.dentally.com


IgnaceMaes (approved):
Looks good!

Thanks for the contribution 😄

PR #1174, opened 2025-09-02
Correct browser versions for 6.x

kategengler:
This was incorrect ... versions are available via require('ember-source').supportedBrowser and can be seen here https://github.com/emberjs/ember.js/blob/main/lib/browsers.js


NullVoxPopuli (approved):

kategengler:
I was able to view but not approve percy

PR #1193, opened 2026-03-25
Update to Ember v4.12

Windvis:
This updates the project to the latest Ember v4.12 blueprint and resolves any new linting errors.

I tried resolving all deprecations I encountered, but some of them are in addon code.

Remaining deprecations:
- ember-data-fastboot triggers the `toArray` deprecation [here](https://github.com/cardstack/ember-data-fastboot/blob/6e6fb8bbf0b405ae174160cc1e4833c5582f68cd/fastboot/instance-initializers/ember-data-fastboot.js#L15). We could maybe patch it? Or inline the code in the repo? The package seems pretty unmaintained. I'm also not really sure we actually need it? Does it provide a huge benefit over fetching the data again in the frontend?
- ember-leaflet triggers a `ember-render-helpers` deprecation, but that shouldn''t block the v5 blueprint update in the future.

Supersedes #1076


app/templates/community/meetups/index.hbs
```diff
@@ -51,7 +51,7 @@
           <ul class="list-unstyled grid lg:grid-4 sm:grid-2">
             {{#each area.meetups as |meetup|}}
               <li>
-                <img width="12" class="align-middle icon-left" src="/images/meetups/map-pin.png" alt="" role="presentation">
+                <img width="12" class="align-middle icon-left" src="/images/meetups/map-pin.png" alt="">
```
void-mAlex: why would we remove role presentation

Windvis: Because it's not needed if `alt=""` is already present. This is something the template-lint autofixer did. ember-template-lint was incorrectly enforcing the `role="presentation"` on elements that also had `alt=""` in older versions.

More info: https://github.com/ember-template-lint/ember-template-lint/pull/2954

tests/integration/helpers/format-date-time-test.js
```diff
@@ -13,6 +13,7 @@ module('Integration | Helper | format-date-time', function (hooks) {
   test('The helper does not error when date is undefined', async function (assert) {
     await render(hbs`
       <div data-test-value>
+        {{! template-lint-disable no-curly-component-invocation }}
         {{format-date-time}}
```
void-mAlex:
```suggestion
       {{(format-date-time)}}
```
isn't it easier/better to just fix the invocation?

Windvis: That doesn't work in a non-Embroidered app. It fails the tests when I try here at least. I could pass in `undefined` explicitly if you think it's better than the disable comment? I didn't really want to mess with that so just went with an ignore comment before.

void-mAlex (approved):

Windvis:
I created a follow-up PR that simply removes ember-data-fastboot: https://github.com/ember-learn/ember-website/pull/1194

That seems like the easiest and imo best solution. I don't think loading an extra json on the initial page load is that bad and it allows us to drop an addon that is not maintained, or having to maintain that code here.

Since the end result of this repo is most likely a Vite + SSR setup it would have to be dropped eventually anyway.

mansona:
I have simplified this PR by splitting out the lint fixes into their own PR https://github.com/ember-learn/ember-website/pull/1197 to make it easier for me to review 👍

mansona (approved):

